PR #5, opened 2015-01-12
clustering: blocking for clustering estimators

glouppe:
This PR implements blocking on top of clustering estimators.

* [x] Implementation of blocking
* [x] Tests for blocking
* [x] Docstrings for blocking

### DIFF
--- a/beard/clustering/blocking.py
+++ b/beard/clustering/blocking.py
@@ -50,7 +50,7 @@ class BlockClustering(BaseEstimator, ClusterMixin):
             The blocking strategy, for mapping samples X to blocks.
             - "single": group all samples X[i] into the same block;
             - "precomputed": use `blocks[i]` argument (in `fit`, `partial_fit`
-              or `predict`) a key for mapping sample X[i] to a block;
+              or `predict`) as a key for mapping sample X[i] to a block;
             - callable: use blocking(X)[i] as a key for mapping sample X[i] to
               a block.
 
@@ -126,7 +126,7 @@ class BlockClustering(BaseEstimator, ClusterMixin):
 
         Returns
         -------
-        :param self: self
+        :returns: self
         """
         # Validate parameters
         X, blocks = self._validate(X, blocks)
@@ -153,7 +153,7 @@ class BlockClustering(BaseEstimator, ClusterMixin):
 
         Returns
         -------
-        :param self: self
+        :returns: self
         """
         # Validate parameters
         X, blocks = self._validate(X, blocks)
@@ -179,7 +179,7 @@ class BlockClustering(BaseEstimator, ClusterMixin):
 
         Returns
         -------
-        :param labels: array-like, shape (n_samples)
+        :returns: array-like, shape (n_samples)
             The labels.
         """
         # Validate parameters
--- a/beard/clustering/blocking.py
+++ b/beard/clustering/blocking.py
@@ -1,0 +1,204 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Beard.
+# Copyright (C) 2014 CERN.
+#
+# Beard is a free software; you can redistribute it and/or modify it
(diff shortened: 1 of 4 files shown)


glouppe:
This is ready for review. @MSusik @etzemis @petros-ioannidis Have some time to take a look?

beard/clustering/blocking.py
```diff
@@ -0,0 +1,204 @@
...
+    labels_ : ndarray, shape (n_samples,)
+        Array of labels assigned to the input data.
+        if partial_fit is used instead of fit, they are assigned to the
+        last batch of data.
+    """
+
+    def __init__(self, blocking="single", base_estimator=None):
+        """Initialize.
+
+        Parameters
```
jirikuncar: IMHO you don't have to add this section before params.

beard/clustering/blocking.py
```diff
@@ -0,0 +1,204 @@
...
+        ----------
+        :param X: {array-like, sparse matrix}, shape (n_samples, n_features)
+            Input data.
+
+        :param blocks: array-like, shape (n_samples, )
+            Block labels, if `blocking == 'precomputed'`.
+
+        Returns
+        -------
+        :param self: self
```
jirikuncar: `:returns:` ?

beard/clustering/blocking.py
glouppe: Indeed, this is not necessary, but I am here following NumPy / SciPy / Scikit-Learn internal conventions. I find it much easier to understand, instead of having a single unstructured list with no clear and apparent separation between parameters, return values or attributes.

jirikuncar: Sphinx is doing the splitting for you when you generate the docs.

glouppe: Yes indeed, but you may also simply access the docstring from an IPython shell or notebook, where no such formatting is done. In this case, having such structure is really helpful.

beard/clustering/tests/test_blocking.py
```diff
@@ -0,0 +1,137 @@
...
+    assert_equal(len(np.unique(clusterer.labels_)), 4)
+
+    # Custom blocking function
+    X_ids = np.arange(len(X)).reshape((-1, 1))
+
+    def _blocking(X_ids):
+        return y[X_ids.ravel()] <= 1  # block labels into {0,1} and {2,3}
+
+    def _distance(X_ids):
+        return euclidean_distances(X[X_ids.ravel()])
```
MSusik: If `X_ids` is a range, isn't `X` the same as `X[X_ids.ravel()]'?

glouppe: The thing is that you dont know what will be the (index) values in `X_ids`, eg. if you do cross-validation `X_ids` will not contain all samples ids. (I agree though that the example here is a bit pointless, as it merely adds a level of indirection to the data, without doing anything clever.)

beard/clustering/blocking.py
```diff
@@ -0,0 +1,204 @@
...
+import numpy as np
+
+from sklearn.base import BaseEstimator
+from sklearn.base import ClusterMixin
+from sklearn.base import clone
+from sklearn.utils import column_or_1d
+
+
+def _single(X):
+    return np.ones(len(X), dtype=np.int)
```
MSusik: If your intention was to reuse this function in future, shouldn't it be moved to a utils file?

glouppe: This is a function specific to blocking, I think it makes more sense to keep it near its usage.

MSusik:
+1 from me.

glouppe:
Thanks for the review @MSusik!

@petros-ioannidis @etzemis would great if you could have a look :)

glouppe:
Since @etzemis is busy until Friday because of the Python course, I am merging this. I need them in order to proceed with the disambiguation example.

PR #9, opened 2015-01-14
examples: simplified author disambiguation

glouppe:
Work-in-progress for a simplified author disambiguation example. 

This example will basically be factored out of https://gist.github.com/glouppe/edaee2c0f82745e301de

* [x] Semi-supervised learning for clustering
* [x] Disambiguation example
* [x] Better affinity function for signatures
* [x] Narrative documentation of the example
* [x] Add tests for `beard.utils.strings`


glouppe:
I have just to add a few tests for `beard.utils.strings`, but otherwise this thing is ready for reviews.

glouppe:
CC @etzemis @petros-ioannidis @MSusik Please feel free to experiment yourself with the example. Suggestions are welcome.

beard/clustering/blocking.py
```diff
@@ -84,41 +97,54 @@ def _validate(self, X, blocks):
...
+            if self.fit_ or not hasattr(clusterer, "partial_fit"):
+                try:
+                    clusterer.fit(X_mask, y=y_mask)
+                except:
+                    clusterer.fit(X_mask)
+            elif self.partial_fit_:
+                try:
+                    clusterer.partial_fit(X_mask, y=y_mask)
+                except:
+                    clusterer.partial_fit(X_mask)
```
MSusik: Isn't it too ignoring?

If there is an exception in `fit` method which is not caused by the lack of `y_mask` we will see an error (or not) from the second method. If you catch only the proper exception (not sure which one will happen here), we can omit the problem.

The same applies two lines below.

beard/clustering/blocking.py
```diff
@@ -84,41 +97,54 @@ def _validate(self, X, blocks):
...
 
-                if hasattr(cluster, "partial_fit"):
-                    cluster.partial_fit(X_mask)
-                else:
-                    cluster.fit(X_mask)
+            if self.fit_ or not hasattr(clusterer, "partial_fit"):
+                try:
+                    clusterer.fit(X_mask, y=y_mask)
+                except:
+                    clusterer.fit(X_mask)
```
MSusik: Isn't it too ignoring?

If there is an exception in `fit` method which is not caused by the lack of `y_mask` we will see an error (or not) from the second method. If you catch only the proper exception (not sure which one will happen here), we can omit the problem.

The same applies two lines below.

beard/clustering/wrappers.py
```diff
@@ -116,6 +127,37 @@ def fit(self, X, y=None):
...
+            best_score = -np.inf
+
+            thresholds = np.concatenate(([0],
+                                         self.linkage_[:, 2],
+                                         [self.linkage_[-1, 2]]))
+
+            for i in range(len(thresholds) - 1):
+                t1, t2 = thresholds[i:i + 2]
+                threshold = (t1 + t2) / 2.0
+                labels = hac.fcluster(self.linkage_, (t1 + t2) / 2.0,
```
MSusik: Probably you wanted to use `threshold` here.

beard/clustering/blocking.py
```diff
@@ -84,41 +97,54 @@ def _validate(self, X, blocks):
...
-                cluster = clone(self.base_estimator)
-                cluster.fit(X_mask)
-
+            # Fit a clusterer on the selected data
+            if len(X_mask) == 1:
+                clusterer = _SingleClustering()
+            elif self.fit_:
+                clusterer = clone(self.base_estimator)
             elif self.partial_fit_:
                 if b in self.clusterers_:
```
MSusik: If we are iteratin over the result `np.unique`, can `b in self.clusteres_` be true?

examples/author_disambiguation.py
```diff
@@ -0,0 +1,162 @@
...
+    return name
+
+
+@memoize
+def initials(name):
+    """Compute the set of initials of a given name."""
+    return set([w[0] for w in name.split()])
+
+
+def affinity(X):
```
MSusik: I suppose that the integration of this function into beard will come in next PR. Am I right?

glouppe: Yes, possibly. Right now this function is really a heuristic which is specific to the data of this very example. This is why I leave it here. In the midterm however, we will have helpers to directly learn such functions.

examples/author_disambiguation.py
```diff
@@ -0,0 +1,162 @@
...
+
+        # We dont know
+        else:
+            distances[i, j] = 0.5
+
+    distances += distances.T
+    return distances
+
+
+def blocking(X):
```
MSusik: The same as above?

glouppe: Yes

glouppe:
Added last tests for `asciify`. This is now ready.

MSusik:
When the test pass, I will give you an :+1:

glouppe:
Test now pass. Thanks @MSusik for the review!  (This `asciify function could be made more robust though...)

glouppe:
@etzemis @petros-ioannidis Could one of you review as well?

etzemis:
Reviewed. No objections :+1:

glouppe:
Great! Thanks @etzemis

PR #12, opened 2015-01-20
clustering: parallelization of blocking

MSusik:
* Parallelizes `_fit` function in BlockClustering class.
  Each job takes care only about specific blocks.

* Introduces usage of joblib library.

Time performance (used `time.time` due to the usage of `multiprocessing`. `time.clock` measures only the time consumed by the main thread):
https://docs.google.com/spreadsheets/d/1mxxvpCxuwPhY4sVlk6wJL2copjNXsV34eMDOzBSFwPQ/edit?usp=sharing

TO DO/TO CONSIDER:
- [x] proper profiling. (not possible)
- [X] try dividing the input data before sending it to the job.  (implemented and attached)
- [X] test for number of jobs more than 1.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,8 @@ matrix:
 install: source travis-install.sh
 
 script:
-    - python setup.py test
+    - python setup.py install
+    - py.test
     - pep257 beard
     - "sphinx-build -qnNW doc doc/_build/html"
 
--- a/beard/clustering/blocking.py
+++ b/beard/clustering/blocking.py
@@ -10,13 +10,17 @@
 """Blocking for clustering estimators.
 
 .. codeauthor:: Gilles Louppe <g.louppe@cern.ch>
+.. codeauthor:: Mateusz Susik <mateusz.susik@cern.ch>
 
 """
 import numpy as np
 
+from joblib import delayed
+from joblib import Parallel
+
 from sklearn.base import BaseEstimator
-from sklearn.base import ClusterMixin
 from sklearn.base import clone
+from sklearn.base import ClusterMixin
 from sklearn.utils import column_or_1d
 
 
@@ -37,6 +41,22 @@ class _SingleClustering(BaseEstimator, ClusterMixin):
         return _single(X)
 
 
+def _parallel_fit(fit_, partial_fit_, b, X, y, clusterer):
+    """Run clusterer's fit function."""
+    if fit_ or not hasattr(clusterer, "partial_fit"):
+        try:
+            clusterer.fit(X, y=y)
+        except TypeError:
+            clusterer.fit(X)
+    elif partial_fit_:
+        try:
+            clusterer.partial_fit(X, y=y)
+        except TypeError:
+            clusterer.partial_fit(X)
+
+    return (b, clusterer)
+
+
 class BlockClustering(BaseEstimator, ClusterMixin):
 
     """Implements blocking for clustering estimators.
@@ -54,7 +74,8 @@ class BlockClustering(BaseEstimator, ClusterMixin):
         last batch of data.
     """
 
-    def __init__(self, affinity=None, blocking="single", base_estimator=None):
+    def __init__(self, affinity=None, blocking="single", base_estimator=None,
+                 n_jobs=1):
         """Initialize.
 
         Parameters
@@ -73,10 +94,14 @@ class BlockClustering(BaseEstimator, ClusterMixin):
 
         :param base_estimator: estimator
             Clustering estimator to fit within each block.
+
+        :param n_jobs: integer
+            Parameter passed directly to joblib library.
         """
         self.affinity = affinity
         self.blocking = blocking
         self.base_estimator = base_estimator
+        self.n_jobs = n_jobs
 
     def _validate(self, X, blocks):
         """Validate hyper-parameters and input data. """
@@ -97,28 +122,38 @@ class BlockClustering(BaseEstimator, ClusterMixin):
 
         return X, blocks
 
-    def _fit(self, X, y, blocks):
-        """Fit base clustering estimators on X."""
-        self.labels_ = -np.ones(len(X), dtype=np.int)
-        offset = 0
+    def _blocks(self, X, y, blocks):
+        """Chop the training data into smaller chunks.
 
-        for b in np.unique(blocks):
-            # Select data from block
+        A chunk is demarcated by the corresponding block. Each chunk contains
+        only the training examples relevant to given block and a clusterer
+        which will be used to fit the data.
+
+        Returns
+        -------
+        :returns: generator
+            Quadruples in the form of ``(block, X, y, clusterer)`` where
+            X and y are the training examples for given block and clusterer is
+            an object with a ``fit`` method.
+        """
+        unique_blocks = np.unique(blocks)
+
+        for b in unique_blocks:
             mask = (blocks == b)
             X_mask = X[mask, :]
-
             if y is not None:
                 y_mask = y[mask]
             else:
                 y_mask = None
-
             if self.affinity == "precomputed":
                 X_mask = X_mask[:, mask]
 
-            # Fit a clusterer on the selected data
+            # Select a clusterer
             if len(X_mask) == 1:
                 clusterer = _SingleClustering()
             elif self.fit_:
+                # Although every job has its own copy of the estimator, one job
+                # can serve multiple fits. That's why the clone is needed.
                 clusterer = clone(self.base_estimator)
             elif self.partial_fit_:
                 if b in self.clusterers_:
@@ -126,24 +161,25 @@ class BlockClustering(BaseEstimator, ClusterMixin):
                 else:
                     clusterer = clone(self.base_estimator)
 
-            if self.fit_ or not hasattr(clusterer, "partial_fit"):
-                try:
-                    clusterer.fit(X_mask, y=y_mask)
-                except TypeError:
-                    clusterer.fit(X_mask)
-            elif self.partial_fit_:
-                try:
-                    clusterer.partial_fit(X_mask, y=y_mask)
-                except TypeError:
-                    clusterer.partial_fit(X_mask)
+            yield (b, X_mask, y_mask, clusterer)
 
-            self.clusterers_[b] = clusterer
+    def _fit(self, X, y, blocks):
+        """Fit base clustering estimators on X."""
+        self.labels_ = -np.ones(len(X), dtype=np.int)
+        offset = 0
 
+        results = (Parallel(n_jobs=self.n_jobs)
+                   (delayed(_parallel_fit)(self.fit_, self.partial_fit_,
+                                           b, X_mask, y_mask, clusterer) for
+                   b, X_mask, y_mask, clusterer in self._blocks(X, y, blocks)))
+
+        for b, clusterer in results:
             # Save predictions
+            self.clusterers_[b] = clusterer
             pred = np.array(clusterer.labels_)
             mask_unknown = (pred == -1)
             pred[~mask_unknown] += offset
-            self.labels_[mask] = pred
+            self.labels_[(blocks == b)] = pred
             offset += np.max(clusterer.labels_) + 1
 
         return self
--- a/beard/clustering/tests/test_blocking.py
+++ b/beard/clustering/tests/test_blocking.py
@@ -17,6 +17,8 @@ from __future__ import division
 import numpy as np
 from numpy.testing import assert_equal
 from numpy.testing import assert_array_equal
+
+from pytest import mark
 import pytest
 
 from sklearn.cluster import AgglomerativeClustering
@@ -29,17 +31,23 @@ from ..blocking import BlockClustering
 from ..wrappers import ScipyHierarchicalClustering
 from ...metrics import paired_f_score
 
+random_state = check_random_state(42)
+X, y = make_blobs(centers=4, shuffle=False, random_state=random_state)
 
-def test_fit():
+
+def _distance(X_ids):
+    return euclidean_distances(X[X_ids.ravel()])
+
+
+@mark.parametrize('n_jobs', (1, 2))
+def test_fit(n_jobs):
     """Test fit."""
-    random_state = check_random_state(42)
-    X, y = make_blobs(centers=4, shuffle=False, random_state=random_state)
-
     # Single block
     clusterer = BlockClustering(
         blocking="single",
         base_estimator=AgglomerativeClustering(n_clusters=4,
-                                               linkage="complete"))
+                                               linkage="complete"),
+        n_jobs=n_jobs)
     clusterer.fit(X)
 
     assert_equal(len(clusterer.clusterers_), 1)
@@ -49,7 +57,8 @@ def test_fit():
     clusterer = BlockClustering(
         blocking="precomputed",
         base_estimator=AgglomerativeClustering(n_clusters=2,
-                                               linkage="complete"))
+                                               linkage="complete"),
+        n_jobs=n_jobs)
     clusterer.fit(X, blocks=(y <= 1))
 
     assert_equal(len(clusterer.clusterers_), 2)
@@ -61,7 +70,8 @@ def test_fit():
         blocking="precomputed",
         base_estimator=ScipyHierarchicalClustering(affinity="precomputed",
                                                    n_clusters=2,
-                                                   method="complete"))
+                                                   method="complete"),
+        n_jobs=n_jobs)
     X_affinity = euclidean_distances(X)
     clusterer.fit(X_affinity, blocks=(y <= 1))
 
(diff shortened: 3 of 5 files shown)


beard/clustering/blocking.py
```diff
@@ -97,58 +150,41 @@ def _validate(self, X, blocks):
...
+                pred = np.array(clusterer.labels_)
+                mask_unknown = (pred == -1)
+                pred[~mask_unknown] += offset
+                self.labels_[(blocks == b)] = pred
+                offset += np.max(clusterer.labels_) + 1
 
         return self
 
-    def fit(self, X, y=None, blocks=None):
+    def fit(self, X, y=None, blocks=None, n_jobs=-1):
```
glouppe: Can you instead make `n_jobs` a parameter in the constructor? (and set it to `1` by default)

API-wise, the idea to have set all hyper-parameters at initialization. `fit` and other methods should only take input data as inputs -- i.e., nothing regarding how the execution is configured.

beard/clustering/blocking.py
```diff
@@ -176,9 +212,9 @@ def fit(self, X, y=None, blocks=None):
         self.clusterers_ = {}
         self.fit_, self.partial_fit_ = True, False
 
-        return self._fit(X, y, blocks)
+        return self._fit(X, y, blocks, n_jobs)
 
-    def partial_fit(self, X, y=None, blocks=None):
+    def partial_fit(self, X, y=None, blocks=None, n_jobs=-1):
```
glouppe: Same here

beard/clustering/blocking.py
```diff
@@ -97,58 +150,41 @@ def _validate(self, X, blocks):
 
         return X, blocks
 
-    def _fit(self, X, y, blocks):
+    def _fit(self, X, y, blocks, n_jobs):
```
glouppe: And here

beard/clustering/tests/test_blocking.py
```diff
@@ -29,12 +29,16 @@
 from ..wrappers import ScipyHierarchicalClustering
 from ...metrics import paired_f_score
 
+random_state = check_random_state(42)
+X, y = make_blobs(centers=4, shuffle=False, random_state=random_state)
+
+
+def _distance(X_ids):
+    return euclidean_distances(X[X_ids.ravel()])
```
MSusik: If we parallelize it with `multiprocessing`, the `distance` function has to be globally accessible.

To work correctly, it needs an access to `X` variable. As many other tests are using the same data I decided to put the definition here. With this, we also test whether the data is not changed by our library.

beard/utils/joblib_helpers.py
```diff
@@ -0,0 +1,34 @@
...
+    :param n_jobs: integer
+        The value used by Joblib library
+
+    Returns
+    -------
+    :returns: integer
+        The number of jobs which would be created if ``n_jobs`` was used
+        by Joblib.
+    """
+    return cpu_count() + 1 + n_jobs if n_jobs == -1 else n_jobs
```
MSusik: Will be changed to
` return cpu_count() + 1 + n_jobs if n_jobs < 0 else n_jobs `

setup.py
```diff
@@ -81,6 +81,7 @@ def run_tests(self):
 
 _install_requires = [
     "scipy>=0.14",
+    "joblib>=0.8.4",
```
glouppe: Add the dependency in `travis-install.sh` as well, to speed up tests.

beard/utils/joblib_helpers.py
```diff
@@ -0,0 +1,34 @@
+# -*- coding: utf-8 -*-
```
glouppe: Can you rename the file to `joblib.py`? (the `_helpers` suffix is redundant with the name of the module)

beard/utils/joblib_helpers.py
```diff
@@ -0,0 +1,34 @@
...
+
+.. codeauthor:: Mateusz Susik <mateusz.susik@cern.ch>
+
+"""
+
+from joblib import cpu_count
+
+
+def jobs_from_njobs(n_jobs):
+    """Compute number of jobs from Joblib's n_jobs variable.
```
glouppe: This could be written in a more clearer way. We should mention the difference between the given value and the output value, i.e. the fact that the former can be negative while the second is always positive.

beard/utils/joblib_helpers.py
```diff
@@ -0,0 +1,34 @@
...
+
+    Parameters
+    ----------
+
+    :param n_jobs: integer
+        The value used by Joblib library
```
glouppe: Same here

beard/utils/joblib_helpers.py
```diff
@@ -0,0 +1,34 @@
...
+"""Joblib library utils.
+
+.. codeauthor:: Mateusz Susik <mateusz.susik@cern.ch>
+
+"""
+
+from joblib import cpu_count
+
+
+def jobs_from_njobs(n_jobs):
```
glouppe: The same goes for the name of the function. I would rather go for something like `positive_n_jobs(n_jobs)`

beard/clustering/blocking.py
```diff
@@ -97,58 +151,41 @@ def _validate(self, X, blocks):
...
         """Fit base clustering estimators on X."""
+        def _grouper(size, seq):
+            return (seq[pos:pos + size] for pos in xrange(0, len(seq), size))
```
glouppe: You should not use `xrange`. It does not save much and it is not Python3 compatible.

beard/clustering/blocking.py
```diff
@@ -37,6 +43,54 @@ def predict(self, X):
         return _single(X)
 
 
+def _parallel_fit(X, y, blocks, affinity, fit_, partial_fit_, base_estimator,
+                  clusterers_, bs):
+
+    res = []
```
glouppe: Can you rename this to `clusterers` and make it a dictionary? This way we would mimic the interface of `clusterers_`.

beard/clustering/blocking.py
```diff
@@ -37,6 +43,54 @@ def predict(self, X):
         return _single(X)
 
 
+def _parallel_fit(X, y, blocks, affinity, fit_, partial_fit_, base_estimator,
+                  clusterers_, bs):
```
glouppe: Can you rename `bs` to something more meaningful?

glouppe:
General question: have you tried building one job/task per block and call the delayed function on the masked data only?

beard/clustering/blocking.py
```diff
@@ -97,58 +151,41 @@ def _validate(self, X, blocks):
...
-
-            # Save predictions
-            pred = np.array(clusterer.labels_)
-            mask_unknown = (pred == -1)
-            pred[~mask_unknown] += offset
-            self.labels_[mask] = pred
-            offset += np.max(clusterer.labels_) + 1
+        true_jobs = jobs_from_njobs(n_jobs)
+        unique_blocks = np.unique(blocks)
+        group_size = (len(unique_blocks) + true_jobs) / true_jobs
```
glouppe: This is not correct

MSusik:
> General question: have you tried building one job/task per block and call the delayed function on the masked data only? 

No, it's in todo list.

beard/clustering/blocking.py
```diff
@@ -99,52 +158,38 @@ def _validate(self, X, blocks):
 
     def _fit(self, X, y, blocks):
         """Fit base clustering estimators on X."""
+        def _grouper(jobs, seq):
```
glouppe: I know you like it, but we couldnt we simply directly import `sklearn.ensemble.base._partition_estimators`? This way we would get rid of some more code :) (Less is better)

beard/clustering/blocking.py
```diff
@@ -99,52 +158,38 @@ def _validate(self, X, blocks):
...
-            offset += np.max(clusterer.labels_) + 1
+        parallelizer = Parallel(n_jobs=self.n_jobs)
+        tasks_iterator = (delayed(_parallel_fit)(X, y, blocks, self.affinity,
+                                                 self.fit_,
+                                                 self.partial_fit_,
+                                                 self.base_estimator,
+                                                 self.clusterers_, b)
+                          for b in _grouper(true_jobs, unique_blocks))
+
+        result = parallelizer(tasks_iterator)
```
glouppe: Can you make the three lines above a single statement? `parallelizer` and `tasks_iterator` variables have no real use here.

travis-install.sh
```diff
@@ -37,7 +37,7 @@ conda update --yes conda
 # provided versions
 conda create -n testenv --yes python=$PYTHON_VERSION pip \
     numpy=$NUMPY_VERSION scipy=$SCIPY_VERSION scikit-learn=$SKLEARN_VERSION \
-    pytest pytest-pep8 pytest-cache sphinx
+    joblib pytest pytest-pep8 pytest-cache sphinx
```
glouppe: Hmmm by bad. This makes Travis fail because `joblib` is not within conda repo.

beard/clustering/blocking.py
```diff
@@ -10,15 +10,21 @@
...
+from joblib import delayed
+from joblib import Parallel
+
 from sklearn.base import BaseEstimator
-from sklearn.base import ClusterMixin
 from sklearn.base import clone
+from sklearn.base import ClusterMixin
 from sklearn.utils import column_or_1d
 
+from ..utils.joblib import positive_n_jobs
```
glouppe: Did this disappear?

beard/clustering/blocking.py
MSusik: Sure, the only thing that I don't like is this underscore. Usually it means that the authors don't want it to be imported.

beard/clustering/blocking.py
MSusik: IMO it's clearer, but I can change it.

travis-install.sh
MSusik: Sure, I'm working on it.

beard/clustering/blocking.py
glouppe: Indeed. In this case however, I believe it is fine -- I am the author of `sklearn.ensemble.base._partition_estimators` :)

MSusik:
The Travis won't pass due to no compability with Python3 -> It's a reminder for myself to remove time functions.

beard/clustering/blocking.py
```diff
@@ -97,54 +144,59 @@ def _validate(self, X, blocks):
...
             if y is not None:
                 y_mask = y[mask]
             else:
                 y_mask = None
-
-            if self.affinity == "precomputed":
+            if affinity == "precomputed":
                 X_mask = X_mask[:, mask]
+            chunk.append((single_block, X_mask, y_mask))
+        return chunk
```
glouppe: Can you make this a generator, hence avoiding precomputing all blocks?

beard/clustering/blocking.py
```diff
@@ -97,54 +144,59 @@ def _validate(self, X, blocks):
...
+
+        self.labels_ = -np.ones(len(X), dtype=np.int)
+        offset = 0
+
+        s = time()
+
+        result = (Parallel(n_jobs=self.n_jobs)
+                  (delayed(_parallel_fit)(self.fit_, self.partial_fit_,
+                                          self.base_estimator,
+                                          self.clusterers_, *b) for b in
```
glouppe: Can you unpack the values yielded by `_block`? Better explicit than implicit.

beard/clustering/blocking.py
```diff
@@ -97,54 +144,59 @@ def _validate(self, X, blocks):
 
         return X, blocks
 
-    def _fit(self, X, y, blocks):
-        """Fit base clustering estimators on X."""
-        self.labels_ = -np.ones(len(X), dtype=np.int)
-        offset = 0
+    def _block(self, X, y, blocks, affinity):
```
glouppe: `_blocks`

glouppe: and `affinity` does not need to be passed, as you have it in `self`

beard/clustering/blocking.py
```diff
@@ -97,54 +144,59 @@ def _validate(self, X, blocks):
...
-
-            if self.affinity == "precomputed":
+            if affinity == "precomputed":
                 X_mask = X_mask[:, mask]
+            chunk.append((single_block, X_mask, y_mask))
+        return chunk
+
+    def _fit(self, X, y, blocks):
+        """Fit base clustering estimators on X."""
+
```
glouppe: remove this empty line

beard/clustering/blocking.py
```diff
@@ -97,54 +144,59 @@ def _validate(self, X, blocks):
...
-            self.clusterers_[b] = clusterer
-
-            # Save predictions
-            pred = np.array(clusterer.labels_)
-            mask_unknown = (pred == -1)
-            pred[~mask_unknown] += offset
-            self.labels_[mask] = pred
-            offset += np.max(clusterer.labels_) + 1
+        t = time()
+        print (t - s)
```
glouppe: remove the timers

beard/clustering/blocking.py
```diff
@@ -97,54 +144,59 @@ def _validate(self, X, blocks):
...
+
+    def _fit(self, X, y, blocks):
+        """Fit base clustering estimators on X."""
+
+        self.labels_ = -np.ones(len(X), dtype=np.int)
+        offset = 0
+
+        s = time()
+
+        result = (Parallel(n_jobs=self.n_jobs)
```
glouppe: `results`

MSusik:
>`from pytest.mark import parametrize`

>This makes the code even shorter :)

Can't do this, `parametrize` is a method run by `mark` object. Need to import all of it.

beard/clustering/blocking.py
```diff
@@ -37,6 +43,42 @@ def predict(self, X):
         return _single(X)
 
 
+def _parallel_fit(fit_, partial_fit_, base_estimator,
+                  clusterers_, block_name, X, y):
+
+    clusterers = {}
```
glouppe: now that we create one task per block, there is no need for this container. We can directly return `(block_name, clusterer)`

glouppe:
In general, can you be consistent in how you name an individual block? For now we have `b`, `block_name` and `single_block`. I think we should use `b` everywhere, as it is clear from the context.

beard/clustering/blocking.py
```diff
@@ -97,54 +144,53 @@ def _validate(self, X, blocks):
...
-            self.labels_[mask] = pred
-            offset += np.max(clusterer.labels_) + 1
+        result = (Parallel(n_jobs=self.n_jobs)
+                  (delayed(_parallel_fit)(self.fit_, self.partial_fit_,
+                                          self.base_estimator,
+                                          self.clusterers_, *b) for b in
+                   self._block(X, y, blocks, self.affinity)))
+
+        for bs in result:
+            for b, clusterer in bs.items():
```
glouppe: provided `_parallel_fit` returns a tuple, we can now remove this double loop and do `for b, clusterer in results:`

beard/clustering/blocking.py
```diff
@@ -97,54 +144,53 @@ def _validate(self, X, blocks):
...
-
-            self.clusterers_[b] = clusterer
-
-            # Save predictions
-            pred = np.array(clusterer.labels_)
-            mask_unknown = (pred == -1)
-            pred[~mask_unknown] += offset
-            self.labels_[mask] = pred
-            offset += np.max(clusterer.labels_) + 1
+        result = (Parallel(n_jobs=self.n_jobs)
```
glouppe: `results`

MSusik:
Ready.

beard/clustering/blocking.py
glouppe: Can you fix this last one?

beard/clustering/blocking.py
```diff
@@ -37,6 +41,38 @@ def predict(self, X):
         return _single(X)
 
 
+def _parallel_fit(fit_, partial_fit_, base_estimator,
+                  clusterers_, b, X, y):
+
```
glouppe: Can you remove this empty line?

glouppe:
Thanks for your patience :) 

+1 for merge on my side, feel free to squash everything

MSusik:
@etzemis @petros-ioannidis Please review.

.travis.yml
```diff
@@ -17,7 +17,8 @@ matrix:
 install: source travis-install.sh
 
 script:
```
MSusik: the `python setup.py test` resulted in a Travis error for Python 2.7:

```
 File "/home/travis/miniconda/envs/testenv/lib/python2.7/atexit.py", line 24, in _run_exitfuncs

func(*targs, **kargs)

File "/home/travis/miniconda/envs/testenv/lib/python2.7/site-packages/sklearn/externals/joblib/pool.py", line 535, in <lambda>

atexit.register(lambda: delete_folder(pool_folder))

TypeError: 'NoneType' object is not callable
```

MSusik:
Will be squashed to a single commit after second review.

etzemis:
:+1:  from me. You may squash them.

MSusik:
Done

PR #13, opened 2015-01-20
metrics: jaro distance metric

petros-ioannidis:
* A new file was created that will be used for string distance metrics.
* Substring_occurences was created that returns a generator with the
occurences of a string.
* Jaro matching was implemented to calculate the matching characters
  based on the Jaro metric algorithm
* Tests were developed for the aforementioned features

### DIFF
--- a/beard/metrics/clustering.py
+++ b/beard/metrics/clustering.py
@@ -39,7 +39,7 @@ def paired_precision_recall_fscore(labels_true, labels_pred):
 
     Returns
     -------
-    :return float precision: calculated precission
+    :return float precision: calculated precision
     :return float recall: calculated recall
     :return float harmonic_mean: calculated harmonic_mean
 
@@ -111,7 +111,7 @@ def paired_precision_score(labels_true, labels_pred):
 
     Returns
     -------
-    :return float precision: calculated precission
+    :return float precision: calculated precision
 
     """
     p, _, _ = paired_precision_recall_fscore(labels_true, labels_pred)
--- a/beard/metrics/tests/test_text.py
+++ b/beard/metrics/tests/test_text.py
@@ -1,0 +1,83 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Beard.
+# Copyright (C) 2015 CERN.
+#
+# Beard is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Test text metrics.
+
+.. codeauthor:: Petros Ioannidis <petros.ioannidis91@gmail.com>
+
+"""
+from __future__ import generators
+import pytest
+from pytest import mark
+from ..text import find_all, _jaro_matching, jaro, jaro_winkler
+from numpy.testing import assert_almost_equal
+
+
+@mark.parametrize('s, letter, occur',
+                  (('MARTHA', 'A', (1, 5)),
+                   ('DWAYNE', 'D', (0, )),
+                   ('A', 'A', (0, )),
+                   ('AABAA', 'AA', (0, 3)),
+                   ('ABCD', 'D', (3, ))))
+def test_find_all_normal_string(s, letter, occur):
+    """Test find_all behaviour for average cases."""
+    assert tuple(find_all(s, letter)) == occur
+
+
+@mark.parametrize('s, letter',
+                  (('MARTHA', 'Z'),
+                   ('', 'A')))
+def test_find_all_none_string(s, letter):
+    """Test find_all behaviour for empty cases."""
+    with pytest.raises(StopIteration):
+        assert next(find_all(s, letter))
+
+
+@mark.parametrize('s, letter',
+                  ((set(), 'A'),
+                   (dict(), 'A'),
+                   (int(), 'A'),
+                   (float(), 'A'),
+                   (list(), 'A')))
+def test_find_all_abnormal_string(s, letter):
+    """Test find_all behaviour called with wrong objects."""
+    with pytest.raises(TypeError):
+        next(find_all(s, letter))
+
+
+@mark.parametrize('s1, s2, match',
+                  (('MARTHA', 'MARHTA', (6, 2)),
+                   ('DWAYNE', 'DUANE', (4, 0)),
+                   ('DUANE', 'DWAYNE', (4, 0)),
+                   ('MARHTA', 'MARTHA', (6, 2))))
+def test_jaro_matching(s1, s2, match):
+    """Test jaro_matching behaviour."""
+    assert _jaro_matching(s1, s2) == match
+
+
+@mark.parametrize('s1, s2, match',
+                  (('MARTHA', 'MARHTA', 0.944),
+                   ('DWAYNE', 'DUANE', 0.822),
+                   ('ABCDEFG', 'ABCDEFG', 1.0),
+                   ('', 'ABCDEFG', 0.0),
+                   ('ABCDEFG', 'HIGKLMN', 0.0)))
+def test_jaro(s1, s2, match):
+    """Test jaro_similarity_metric behaviour."""
+    assert_almost_equal(jaro(s1, s2), match, 3)
+
+
+@mark.parametrize('s1, s2, match',
+                  (('MARTHA', 'MARHTA', 0.961),
+                   ('DWAYNE', 'DUANE', 0.84),
+                   ('ABCDEFG', 'ABCDEFG', 1.0),
+                   ('', 'ABCDEFG', 0.0),
+                   ('ABCDEFG', 'HIGKLMN', 0.0)))
+def test_jaro_winkler(s1, s2, match):
+    """Test jaro_similarity_metric behaviour."""
+    assert_almost_equal(jaro_winkler(s1, s2), match, 3)
--- a/beard/metrics/text.py
+++ b/beard/metrics/text.py
@@ -1,0 +1,166 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Beard.
+# Copyright (C) 2015 CERN.
+#
+# Beard is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Text metrics.
+
+.. codeauthor:: Petros Ioannidis <petros.ioannidis91@gmail.com>
+
+"""
+
+from __future__ import division
+import re
+
+
+def find_all(s, pattern):
+    """Find all occurences of the given pattern.
+
+    Parameters
+    ----------
+    :param s: string
+        String to be searched
+
+    :param letter: string
+        Substring we are searching for
+
+    Returns
+    -------
+    :returns: generator
+        A generator that holds the indexes of the patterns
+    """
+    for match in re.finditer(pattern, s):
+        yield match.start()
+
+
+def _jaro_matching(s1, s2):
+    """Return the number of matching letters and transpositions.
+
+    Parameters
+    ----------
+    :param s1: string
+        First string
+
+    :param s2: string
+        Second string
+
+    Returns
+    -------
+    :returns: (int, int)
+        The number of matching letters and transpositions
+    """
+    H = min(len(s1), len(s2))//2
+
+    letters_cache = {}
+    matches = 0
+    transpositions = 0
+    s1_matching_letter = []
+    s2_matching_letter = []
+    s1_matched_positions = []
+    s2_matched_positions = []
+
+    for letter in s1:
+        if letter not in letters_cache:
+            letters_cache[letter] = (tuple(find_all(s1, letter)),
+                                     tuple(find_all(s2, letter)))
+
+    for letter, (s1_positions, s2_positions) in letters_cache.items():
+        for i in s1_positions:
+            for j in s2_positions:
+                if i - H <= j <= i + H:
+                    if j not in s2_matched_positions:
+                        matches += 1
+                        s2_matched_positions.append(j)
+                    s1_matching_letter.append((i, letter))
+                    break
+
+    for letter, (s1_positions, s2_positions) in letters_cache.items():
+        for j in s2_positions:
+            for i in s1_positions:
+                if i - H <= j <= i + H:
+                    if i not in s1_matched_positions:
+                        s1_matched_positions.append(i)
+                    s2_matching_letter.append((j, letter))
+                    break
+
+    s1_matching_letter.sort()
+    s2_matching_letter.sort()
+    transpositions = len(tuple(filter(lambda x: x[0][1] != x[1][1],
+                               zip(s1_matching_letter,
+                                   s2_matching_letter))))
+    return matches, transpositions
+
+
+def jaro(s1, s2):
+    """Return the Jaro similarity of the strings s1 and s2.
+
+    Parameters
+    ----------
+    :param s1: string
+        First string
+
+    :param s2: string
+        Second string
+
+    Returns
+    -------
+    :returns: float
+        Similarity of s1 and s2
+
+    Reference
+    ---------
+    Jaro, M. A. 1989. Advances in record-linkage methodology as
+    applied to matching the 1985 census of Tampa, Florida.
+    Journal of the American Statistical Association
+    84:414-420.
+    """
+    if len(s1) == 0 or len(s2) == 0:
+        return 0
+
+    n_matches, n_transpositions = _jaro_matching(s1, s2)
+
+    if n_matches == 0:
+        return 0
+
+    return (1/3)*((n_matches/len(s1)) +
+                  (n_matches/len(s2)) +
+                  ((n_matches - (n_transpositions/2))/(n_matches)))
+
+
+def jaro_winkler(s1, s2, p=0.1):
+    """Return the Jaro-Winkler similarity of the strings s1 and s2.
+
+    Parameters
+    ----------
+    :param s1: string
+        First string
+
+    :param s2: string
+        Second string
+
+    Returns
+    -------
+    :returns: float
+        Similarity of s1 and s2
+
+    Reference
+    ---------
+    Winkler, W. E. 1999. The state of record linkage and cur-
+    rent research problems. Statistics of Income Division, In-
+    ternal Revenue Service Publication R99/04. Available from
+    http://www.census.gov/srd/www/byname.html.
+    """
+    jaro_distance = jaro(s1, s2)
+
+    common_prefix = 0
+    for s1_letter, s2_letter in zip(s1, s2):
+        if s1_letter == s2_letter and common_prefix < 4:
+            common_prefix += 1
+        else:
+            break
+
+    return jaro_distance + (p * common_prefix * (1 - jaro_distance))


beard/metrics/string_distance.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Beard.
+# Copyright (C) 2014, 2015 CERN.
```
glouppe: 2015 only

beard/metrics/string_distance.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
```
glouppe: Can you rename this file to simply `string.py`? The `_distance` suffix is redundant with the fact that the file is already within a folder named "metrics".

beard/metrics/string_distance.py
```diff
@@ -0,0 +1,75 @@
...
+#
+# This file is part of Beard.
+# Copyright (C) 2014, 2015 CERN.
+#
+# Beard is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""String distances metrics.
```
glouppe: Remove the "distances" term, as this is not strictly correct.

beard/metrics/string_distance.py
```diff
@@ -0,0 +1,75 @@
...
+        substring we are searching for
+
+    Returns
+    -------
+    :returns: generator
+        a generator that holds the value of the index
+    """
+    if letter == '':
+        raise ValueError('Empty string is not allowed as a parameter for letter')
+    try:
```
glouppe: I would remove this `try/except` and directly start from `index = 0` in the loop below. There is no need to check that `S` supports `find` (garbage in, garbage out).

beard/metrics/string_distance.py
```diff
@@ -0,0 +1,75 @@
...
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""String distances metrics.
+
+.. codeauthor:: Petros Ioannidis <petros.ioannidis91@gmail.com>
+
+"""
+def substring_occurences(s, letter):
+    """Return the number of matching letters based on the jaro algorithm.
```
glouppe: I would rephrase this as 
```
def find_all(s, pattern):
    """Find all occurences of the given pattern."""
```

beard/metrics/string_distance.py
```diff
@@ -0,0 +1,75 @@
...
+    if letter == '':
+        raise ValueError('Empty string is not allowed as a parameter for letter')
+    try:
+        index = s.find(letter)
+    except AttributeError:
+        raise TypeError('S must be a string')
+
+    while index != -1:
+        yield index
+        index = s.find(letter, index + 1)
```
glouppe: By the way, couldnt we simply do instead : 

```
for match in re.finditer(pattern, s):
    yield match.start()
```

beard/metrics/string_distance.py
```diff
@@ -0,0 +1,75 @@
...
+
+def jaro_matching(s1, s2):
+    """Return the number of matching letters based on the jaro algorithm.
```
glouppe: I would rephrase this as "Return the number of matching letters in s1 and s2".

beard/metrics/tests/test_string_distance.py
```diff
@@ -0,0 +1,56 @@
...
+"""Test string distances metrics.
+
+.. codeauthor:: Petros Ioannidis <petros.ioannidis91@gmail.com>
+
+"""
+import pytest
+from ..string_distance import substring_occurences, jaro_matching
+
+
+@pytest.mark.parametrize('s, letter, occur',
```
glouppe: Can you import `parametrize` above to remove some boilerplate code?

glouppe:
Since this is still quite short, I will wait for you to implement the full Jaro distance function before merging this.

beard/metrics/tests/test_string_distance.py
```diff
@@ -0,0 +1,56 @@
+# -*- coding: utf-8 -*-
```
glouppe: Same here, rename the file to `test_string.py`

beard/metrics/tests/test_string_distance.py
```diff
@@ -0,0 +1,56 @@
...
+# # This file is part of Beard.
+# Copyright (C) 2014, 2015 CERN.
```
glouppe: 2015 only

beard/metrics/tests/test_string_distance.py
```diff
@@ -0,0 +1,56 @@
...
+#
+# Beard is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Test string distances metrics.
```
glouppe: without "distances"

beard/metrics/string_distance.py
petros-ioannidis: There is a `string` package in python so I suggest we try to find another name.

beard/metrics/string_distance.py
petros-ioannidis: You are right. I runned timeit and the code you proposed runs faster. I will change it.

beard/metrics/string_distance.py
petros-ioannidis: I used `string_methods.py` for a temporary name until we find something more meaningful.

beard/metrics/string_distance.py
```diff
@@ -0,0 +1,75 @@
...
+        The number of matching letters
+    """
+
+    H = min(len(s1), len(s2))//2
+
+    letters_cache = dict()
+    matches = 0
+
+    for letter in s1:
+        if not letters_cache.has_key(letter):
```
MSusik: `in` operator does the same, is faster and is easier to read.

beard/metrics/string_distance.py
```diff
@@ -0,0 +1,75 @@
...
+    :param s1: string
+        first string
+
+    :param s2: string
+        second string
+
+    Returns
+    -------
+    :returns: int
+        The number of matching letters
```
MSusik: You have some inconsistency in your comments:

` a generator that holds the value of the index`
` The number of matching letters`
uppercase vs lowercase, "a" vs "the"

The trend is to use uppercase and omit articles. Maybe we should standarize it.

petros-ioannidis: Omitting articles is not suggested by PEP8. I fixed the uppercase.

beard/metrics/string_methods.py
```diff
@@ -0,0 +1,71 @@
...
+    :param letter: string
+        Substring we are searching for
+
+    Returns
+    -------
+    :returns: generator
+        A generator that holds the indexes of the patterns
+    """
+    for match in re.finditer(pattern, s):
+            yield match.start()
```
glouppe: indentation is not consistent here

beard/metrics/string_methods.py
```diff
@@ -0,0 +1,71 @@
...
+
+def jaro_matching(s1, s2):
+    """Return the number of matching letters based on the jaro algorithm.
```
glouppe: Strictly speaking, there is no "jaro algorithm", but only a "jaro similarity metric" . So I would rephrase this as "Return the number of matching letters in the two strings"

glouppe:
For the filename, `text.py` should be fine.

beard/metrics/tests/test_string_methods.py
```diff
@@ -0,0 +1,53 @@
...
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Test string metrics.
+
+.. codeauthor:: Petros Ioannidis <petros.ioannidis91@gmail.com>
+
+"""
+import pytest
+from pytest import mark
```
glouppe: Why not `from pytest.mark import parametrize`?

petros-ioannidis: `mark` is not a module.

glouppe:
@petros-ioannidis During development, can you avoid amending/squashing commits? This makes much harder to follow the discussion in github. We will squash everything before merge anyway.

beard/metrics/tests/test_string_methods.py
glouppe: Ok, I just realized it. (yuck)

beard/metrics/text.py
```diff
@@ -0,0 +1,119 @@
...
+    Returns
+    -------
+    :returns: generator
+        A generator that holds the indexes of the patterns
+    """
+    for match in re.finditer(pattern, s):
+        yield match.start()
+
+
+def jaro_matching(s1, s2):
```
glouppe: Can you rename this function? Since it now returns both the number of matching letters and number of transpositions. Please update the docstring as well.

beard/metrics/text.py
```diff
@@ -0,0 +1,119 @@
...
+                    break
+    s1_matching_letter_sequence.sort()
+    s2_matching_letter_sequence.sort()
+    transpositions = len(filter(lambda x: x[0][1] != x[1][1],
+                         zip(s1_matching_letter_sequence,
+                             s2_matching_letter_sequence)))
+    return matches, transpositions
+
+
+def jaro_similarity_metric(s1, s2):
```
glouppe: Simply name it `jaro`

beard/metrics/text.py
```diff
@@ -0,0 +1,119 @@
...
+        Similarity of s1 and s2
+    """
+    s_prime = jaro_matching(s1, s2)[0]
+    s = len(s1)
+    t_prime = jaro_matching(s2, s1)[0]
+    t = len(s2)
+
+    return (1/3)*((s_prime/s) +
+                  (t_prime/t) +
+                  ((s_prime - (jaro_matching(s1, s2)[1]/2))/(s_prime)))
```
glouppe: What definition of the metric are you using?

beard/metrics/text.py
petros-ioannidis: Yes I forgot. Should I make 2 different functions or a tuple is fine? I did it like this for 'optimization' purposes but maybe its less clear like that. What do you think?

glouppe: Computing the two values at the same time is fine with me.

beard/metrics/clustering.py
```diff
@@ -38,7 +38,7 @@ def paired_precision_recall_fscore(labels_true, labels_pred):
 
     Returns
     -------
-    :return float precision: calculated precission
+    :return float precision: calculated precision
```
etzemis: This is an outdated version of the file.

petros-ioannidis: I rebased, it should be fine now.

beard/metrics/text.py
glouppe: In the end, we dont reuse nor call this function more than once. Shouldn't we put directly the computation of the number of matching letters / transpositions directly into `jaro`?

beard/metrics/text.py
```diff
@@ -0,0 +1,126 @@
...
+    :param s2: string
+        Second string
+
+    Returns
+    -------
+    :returns: float
+        Similarity of s1 and s2
+    """
+    s_data = jaro_matching(s1, s2)
+    s_prime = s_data[0]
```
glouppe: Could you instead unpack both values directly? It would be more readable. 

E.g. `n_matches, n_transpositions = jaro_matching(s1, s2)`

beard/metrics/text.py
```diff
@@ -0,0 +1,126 @@
...
+        Second string
+
+    Returns
+    -------
+    :returns: int
+        The number of matching letters
+    """
+    H = min(len(s1), len(s2))//2
+
+    letters_cache = dict()
```
glouppe: Use `{}`, for consistency with how you initialize lists below.

beard/metrics/text.py
```diff
@@ -0,0 +1,126 @@
...
+                if pos_s1 - H <= pos_s2 <= pos_s1 + H:
+                    if pos_s1 not in s1_matched_positions:
+                        s1_matched_positions.append(pos_s1)
+                    s2_matching_letter_sequence.append((pos_s2, letter))
+                    break
+
+    s1_matching_letter_sequence.sort()
+    s2_matching_letter_sequence.sort()
+    transpositions = len(tuple(filter(lambda x: x[0][1] != x[1][1],
+                         zip(s1_matching_letter_sequence,
```
glouppe: Can you align `zip` with `filter`, to make it more readable?

beard/metrics/text.py
```diff
@@ -0,0 +1,126 @@
...
+                             s2_matching_letter_sequence))))
+    return matches, transpositions
+
+
+def jaro(s1, s2):
+    """Return the similarity of the strings s1 and s2.
```
glouppe: Return the Jaro similarity between strings `s1` and `s2`.

beard/metrics/text.py
```diff
@@ -0,0 +1,126 @@
...
+
+    Calculate the similarity based on the Jaro similarity metric.
```
glouppe: and then remove this sentence, which becomes redundant with the one above

beard/metrics/text.py
```diff
@@ -0,0 +1,126 @@
...
+        for pos_s2 in s2_occur:
+            for pos_s1 in s1_occur:
+                if pos_s1 - H <= pos_s2 <= pos_s1 + H:
+                    if pos_s1 not in s1_matched_positions:
+                        s1_matched_positions.append(pos_s1)
+                    s2_matching_letter_sequence.append((pos_s2, letter))
+                    break
+
+    s1_matching_letter_sequence.sort()
+    s2_matching_letter_sequence.sort()
```
glouppe: Hmm I feel there is a more straightforward way to compute all these values. I am not not sure the caching you do above is worth it in the end because you need to resort the sequences. I'll check this more closely tomorrow.

beard/metrics/text.py
```diff
@@ -0,0 +1,126 @@
...
+    s2_matching_letter_sequence = []
+    s1_matched_positions = []
+    s2_matched_positions = []
+
+    for letter in s1:
+        if letter not in letters_cache:
+            letters_cache[letter] = (tuple(find_all(s1, letter)),
+                                     tuple(find_all(s2, letter)))
+
+    for letter, (s1_occur, s2_occur) in letters_cache.items():
```
MSusik: `items()` has different behaviour in Python2 and Python3. If you want to be consistent, use `six` library.

http://pythonhosted.org/six/#six.iteritems

beard/metrics/text.py
```diff
@@ -0,0 +1,126 @@
...
+        for pos_s1 in s1_occur:
+            for pos_s2 in s2_occur:
+                if pos_s1 - H <= pos_s2 <= pos_s1 + H:
+                    if pos_s2 not in s2_matched_positions:
+                        matches += 1
+                        s2_matched_positions.append(pos_s2)
+                    s1_matching_letter_sequence.append((pos_s1, letter))
+                    break
+
+    for letter, (s1_occur, s2_occur) in letters_cache.items():
```
MSusik: The same as above

beard/metrics/tests/test_text.py
```diff
@@ -0,0 +1,68 @@
...
+    assert tuple(find_all(s, letter)) == occur
+
+
+@mark.parametrize('s, letter',
+                  (('MARTHA', 'Z'),
+                   ('', 'A')))
+def test_find_all_none_string(s, letter):
+    """Test find_all behaviour for empty cases."""
+    with pytest.raises(StopIteration):
+        assert next(find_all(s, letter))
```
MSusik: Might be nice to use here `six` library:

http://docs.pythonsprints.com/python3_porting/py-porting.html#iterators
http://pythonhosted.org/six/#six.next

beard/metrics/text.py
petros-ioannidis: When used as an `iterable` the behaviour is the same in Python2 and Python3. So it is fine like that.

MSusik: IMO, no! As a C++ programmer you should know that the type should be as strict as possible.

Anyway, let's drop the opinionable argument and let's think about the performance.
In python 2 `items()` method creates a new list of tuples. That is both time and memory consuming.
`six.iteritems()` just uses python 2 `iteritems()`.

petros-ioannidis: We are not coding in `C++` though, so I suggest you focus on the pythonic way. ;)
Items is fine for this use case. If you want you can benchmark it and if it is really a blocker for performance I can change it. ;)

beard/metrics/text.py
glouppe: Can you make this private?

petros-ioannidis: Done.

glouppe:
Thanks for the last changes. Can you proceed and implement jaro-winkler in this same branch? (it should be quite straightforward now)

beard/metrics/tests/test_text.py
```diff
@@ -0,0 +1,76 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Beard.
+# Copyright (C) 2015 CERN.
+#
+# Beard is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Test string metrics.
```
glouppe: string -> text (for consistency with the file name)

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Beard.
+# Copyright (C) 2015 CERN.
+#
+# Beard is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""String metrics.
```
glouppe: string -> text

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+    """
+    for match in re.finditer(pattern, s):
+        yield match.start()
+
+
+def _jaro_matching(s1, s2):
+    """Return the number of matching letters.
+
+    Calculate the number of matching letters
+    based on the method for computing the Jaro distance metric.
```
glouppe: Can you remove this sentence? I think it is redundant with the first sentence.

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+    -------
+    :returns: generator
+        A generator that holds the indexes of the patterns
+    """
+    for match in re.finditer(pattern, s):
+        yield match.start()
+
+
+def _jaro_matching(s1, s2):
+    """Return the number of matching letters.
```
glouppe: The function is also computing the number of transpositions. Hence it should be mentioned. 

Accordingly, the function should be renamed as well.

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+    :param s1: string
+        First string
+
+    :param s2: string
+        Second string
+
+    Returns
+    -------
+    :returns: int
+        The number of matching letters
```
glouppe: Same here, please correct the docsting

beard/metrics/text.py
glouppe: Can you rename `s1_occur` to `s1_positions`? I think it would be clearer. The same for `s2_occur`.

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+    s2_matched_positions = []
+
+    for letter in s1:
+        if letter not in letters_cache:
+            letters_cache[letter] = (tuple(find_all(s1, letter)),
+                                     tuple(find_all(s2, letter)))
+
+    for letter, (s1_occur, s2_occur) in letters_cache.items():
+        for pos_s1 in s1_occur:
+            for pos_s2 in s2_occur:
```
glouppe: can you rename `pos_s1` to `i` and `pos_s2` to `j`?

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+    -------
+    :returns: int
+        The number of matching letters
+    """
+    H = min(len(s1), len(s2))//2
+
+    letters_cache = {}
+    matches = 0
+    transpositions = 0
+    s1_matching_letter_sequence = []
```
glouppe: `s1_matching_letters` would be a bit shorter

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+        Second string
+
+    Returns
+    -------
+    :returns: float
+        Similarity of s1 and s2
+    """
+    n_matches, n_transpositions = _jaro_matching(s1, s2)
+    s = len(s1)
+    t = len(s2)
```
glouppe: Can you remove these two variable and directly call `len(s1)` (`len(s2)`) in the expression below?

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+
+    return (1/3)*((n_matches/s) +
+                  (n_matches/t) +
+                  ((n_matches - (n_transpositions/2))/(n_matches)))
+
+
+def jaro_winkler(s1, s2):
```
glouppe: Can you instead make the scaling factor `p` a parameter of this function, with `p=0.1` by default?

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+
+    :param s2: string
+        Second string
+
+    Returns
+    -------
+    :returns: float
+        Similarity of s1 and s2
+    """
+    _jaro = jaro(s1, s2)
```
glouppe: `_jaro` -> `jaro_distance`

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+
+    P = 0
+    for s1_letter, s2_letter in zip(s1, s2):
+        if s1_letter == s2_letter:
+            P += 1
+        else:
+            break
+
+    P_prime = min(P, 4)
```
glouppe: this variable is not useful, directly use `min()` in the expression below

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+        Second string
+
+    Returns
+    -------
+    :returns: float
+        Similarity of s1 and s2
+    """
+    _jaro = jaro(s1, s2)
+
+    P = 0
```
glouppe: I would rename this as `common_prefix` to make it easier to understand.

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+    for s1_letter, s2_letter in zip(s1, s2):
```
glouppe: Note that we could stop the loop early, as soon as `P==4`.

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
+        First string
+
+    :param s2: string
+        Second string
+
+    Returns
+    -------
+    :returns: float
+        Similarity of s1 and s2
+    """
```
glouppe: Can you add a reference towards the original paper?

beard/metrics/text.py
```diff
@@ -0,0 +1,153 @@
...
```
glouppe: Can you add a reference towards the original paper?

beard/metrics/text.py
petros-ioannidis: Wouldn't this alter the algorithm? It is specified in the description of the algorithm.

glouppe: I have seen several papers where they adjust the scaling factor. I think it is best to leave open the possibility to do that as well, as long as the default value matches the default algorithm.

beard/metrics/tests/test_text.py
```diff
@@ -0,0 +1,76 @@
...
+    """Test jaro_matching behaviour."""
+    assert _jaro_matching(s1, s2) == match
+
+
+@mark.parametrize('s1, s2, match',
+                  (('MARTHA', 'MARHTA', 0.944),
+                   ('DWAYNE', 'DUANE', 0.822)))
+def test_jaro(s1, s2, match):
+    """Test jaro_similarity_metric behaviour."""
+    assert abs(jaro(s1, s2) - match) < 0.001
```
glouppe: Can you use assert_almost_equal from numpy.testing instead?

beard/metrics/tests/test_text.py
```diff
@@ -0,0 +1,76 @@
...
+
+
+@mark.parametrize('s1, s2, match',
+                  (('MARTHA', 'MARHTA', 0.961),
+                   ('DWAYNE', 'DUANE', 0.84)))
+def test_jaro_winkler(s1, s2, match):
+    """Test jaro_similarity_metric behaviour."""
+    assert abs(jaro_winkler(s1, s2) - match) < 0.001
```
glouppe: Same here

beard/metrics/text.py
```diff
@@ -0,0 +1,170 @@
...
+    jaro_distance = jaro(s1, s2)
+
+    common_prefix = 0
+    for s1_letter, s2_letter in zip(s1, s2):
+        if s1_letter == s2_letter and common_prefix < 4:
+            common_prefix += 1
+        else:
+            break
+
+    return jaro_distance + (p * min(common_prefix, 4) * (1 - jaro_distance))
```
glouppe: Now that the loop breaks early, you can directly use `common_prefix` instead of `min(...)`

beard/metrics/text.py
```diff
@@ -0,0 +1,170 @@
...
+    -------
+    :returns: float
+        Similarity of s1 and s2
+
+    Reference
+    ---------
+    Jaro, M. A. 1989. Advances in record-linkage methodology as
+    applied to matching the 1985 census of Tampa, Florida.
+    Journal of the American Statistical Association
+    84:414–420.
```
glouppe: This one is enough

beard/metrics/tests/test_text.py
```diff
@@ -0,0 +1,76 @@
...
+                   ('MARHTA', 'MARTHA', (6, 2))))
+def test_jaro_matching(s1, s2, match):
+    """Test jaro_matching behaviour."""
+    assert _jaro_matching(s1, s2) == match
+
+
+@mark.parametrize('s1, s2, match',
+                  (('MARTHA', 'MARHTA', 0.944),
+                   ('DWAYNE', 'DUANE', 0.822)))
+def test_jaro(s1, s2, match):
```
glouppe: Can you test the border cases as well? ie. when strings are identical / dont share any letters / one is empty.

glouppe:
Great, thanks for your patience. +1 for me. Can you squash?

@MSusik is it ok for you?

MSusik:
> @MSusik is it ok for you?

IMO using `items()` is wrong. Ask @etzemis  for a review.

petros-ioannidis:
The commits are squashed and the PR is ready to be merged.

etzemis:
:+1: for me for merging.

glouppe:
Thanks, merging this

PR #16, opened 2015-02-02
clustering: BlockClustering.labels_ as a property

glouppe:
This defines `labels_` as a property, thereby allowing on-the-fly changes if the underlying clusterers are modified. 

* [x] `labels_` as a property
* [x] tests

### DIFF
--- a/beard/clustering/blocking.py
+++ b/beard/clustering/blocking.py
@@ -72,6 +72,9 @@ class BlockClustering(BaseEstimator, ClusterMixin):
         Array of labels assigned to the input data.
         if partial_fit is used instead of fit, they are assigned to the
         last batch of data.
+
+    blocks_ : ndarray, shape (n_samples,)
+        Array of keys mapping input data to blocks.
     """
 
     def __init__(self, affinity=None, blocking="single", base_estimator=None,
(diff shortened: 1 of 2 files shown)


glouppe:
Ready for reviews -- it should not take long! @MSusik @etzemis

beard/clustering/blocking.py
```diff
@@ -276,12 +269,32 @@ def predict(self, X, blocks=None):
...
+        offset = 0
+
+        for b in self.clusterers_:
+            clusterer = self.clusterers_[b]
+
+            pred = np.array(clusterer.labels_)
+            mask_unknown = (pred == -1)
+            pred[~mask_unknown] += offset
+            labels[(self.blocks_ == b)] = pred
+            offset += np.max(clusterer.labels_) + 1
```
MSusik: The code seems to be duplicated.

glouppe: Indeed, it is quite close to the one in `predict`. Yet, the two are doing things that are slightly different, hence I am not sure factoring this piece of code out is worth it with respect to the gain in lines of code. In the meantime, I simplified the logic a bit to make this 'redundant" code shorter. What do you think?

MSusik: OK, it's not that bad. Let 's just remember that when we do the same thing one more time, we should factor it out. :+1:

glouppe:
If you have time @etzemis, I would love a review for this :)

etzemis:
Reviewed. I do not have anything more to mention. Thus :+1: for merging.

PR #18, opened 2015-02-02
metrics: levenshtein distance metric

etzemis:


glouppe:
Can you rebase?

beard/metrics/tests/test_text.py
```diff
@@ -10,12 +10,13 @@
...
 
 .. codeauthor:: Petros Ioannidis <petros.ioannidis91@gmail.com>
+.. codeauthor:: Evangelos Tzemis <evangelos.tzemis@gmail.com>
 
 """
 from __future__ import generators
 import pytest
 from pytest import mark
-from ..text import find_all, _jaro_matching, jaro, jaro_winkler
+from ..text import find_all, _jaro_matching, jaro, jaro_winkler, levenshtein
```
glouppe: Can reorder these imports, from the standard library to local imports. Could you also make one import per line?

beard/metrics/text.py
```diff
@@ -164,3 +166,48 @@ def jaro_winkler(s1, s2, p=0.1):
             break
 
     return jaro_distance + (p * common_prefix * (1 - jaro_distance))
+
+
+def levenshtein(a, b):
+    """Calculate the levenshtein distance between strings a and b.
```
glouppe: Can you add a note about case sensitivity?

beard/metrics/tests/test_text.py
```diff
@@ -81,3 +82,18 @@ def test_jaro(s1, s2, match):
...
+                   ('Adam', 'Adams', 1),
+                   ('YES', 'yes', 3),
+                   ('weight', 'muchweigh', 5),
+                   ('grand father', '', len('grand father')),
+                   ('', 'grand father', len('grand father')),
+                   (' ', ' ', 0),
+                   ('', '', 0)))
+def test_levenshtein(string_a, string_b, cost):
+    """Test levenshtein_metric behaviour."""
+    assert levenshtein(string_a, string_b) == cost
```
glouppe: Can you rename `cost` to `distance` instead? (here and above)

glouppe:
Other than my comments, I am +1 for merge.

glouppe:
Second review from @MSusik or @natsheh

beard/metrics/text.py
```diff
@@ -162,3 +164,51 @@ def jaro_winkler(s1, s2, p=0.1):
             break
 
     return jaro_distance + p * common_prefix * (1 - jaro_distance)
+
+
+def levenshtein(a, b):
+    """Calculate the levenshtein distance between strings a and b.
+
+    Case sensitiveness is activated, meaning that uppercase letters
```
MSusik: Wouldn't it be useful to have case sensitiveness as a parameter in this function?

glouppe: I would prefer to keep a clear separation of concerns. If someone wants case insensitiveness, then this should be explicit by calling `.lower()` before calling this function. But this is just my opinion. If others feel it would be nice, then fine.

natsheh: May be we would consider using this package for Levenshtein dissimilarity: https://pypi.python.org/pypi/python-Levenshtein/0.12.0 or at least check if ours is faster

MSusik: > I would prefer to keep a clear separation of concerns.

OK, it's fine for me. :+1: from me.

PR #19, opened 2015-02-04
similarity: transformers for paired data

glouppe:
This PR implements transformers for paired data.

* [x] Generic transformers
* [x] Paired data transformers
* [x] Documentation
* [x] Tests

### DIFF
--- a/beard/similarity/pairs.py
+++ b/beard/similarity/pairs.py
@@ -1,0 +1,212 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Beard.
+# Copyright (C) 2015 CERN.
+#
+# Beard is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Transformers for paired data.
+
+.. codeauthor:: Gilles Louppe <g.louppe@cern.ch>
+
+"""
+import numpy as np
+import scipy.sparse as sp
+
+from sklearn.base import TransformerMixin
+
+
+class PairTransformer(TransformerMixin):
+
+    """Apply a transformer on all elements in paired data."""
+
+    def __init__(self, element_transformer):
+        """Initialize.
+
+        Parameters
+        ----------
+        :param element_transformer: transformer
+            The transformer to apply on each element.
+        """
+        self.element_transformer = element_transformer
+
+    def fit(self, X, y=None):
+        """Fit the given transformer on all individual elements in `X`.
+
+        Rows i in `X` are assumed to represent pairs, where
+        `X[i, :n_features]` and `X[i, n_features:]` correspond to their two
+        individual elements. Calling `fit` trains the given transformer on
+        the dataset formed by all these individual elements.
+
+        Parameters
+        ----------
+        :param X: array-like, shape (n_samples, 2 * n_features)
+            Input paired data.
+
+        Returns
+        -------
+        :returns: self
+        """
+        n_samples, n_features_all = X.shape
+        self.n_features_ = n_features_all / 2
+
+        if sp.issparse(X):
+            X_ = sp.vstack((X[:, :self.n_features_],
+                            X[:, self.n_features_:])).tocsr()
+        else:
+            X_ = np.vstack((X[:, :self.n_features_], X[:, self.n_features_:]))
+
+        self.element_transformer.fit(X_)
+
+        return self
+
+    def transform(self, X):
+        """Transform all individual elements in `X`.
+
+        Rows i in the returned array `X_` represent transformed pairs, where
+        `X_[i, :n_features_prime]` and `X_[i, n_features_prime:]` correspond
+        to their two individual transformed elements.
+
+        Parameters
+        ----------
+        :param X: array-like, shape (n_samples, 2 * n_features)
+            Input paired data.
+
+        Returns
+        -------
+        :returns X_: array-like, shape (n_samples, 2 * n_features_prime)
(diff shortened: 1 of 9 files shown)


glouppe:
@etzemis You may want to have a look at this, to get inspiration for implementing "soft" similarities.

glouppe:
This is ready for reviews! Basically these classes have been extracted from the prototype in the notebook. 

(Then the next step will be to reuse all of this to illustrate an advanced use case of author disambiguation (i.e., plugging distance learning + block clustering)). 

CC: @MSusik @etzemis @natsheh

glouppe:
This PR now also includes `utils.normalize_personal_name`. This latter function might need some more work to make it more robust (along with `asciify`), but this can be done later in a separate PR.

beard/similarity/pairs.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        return self
+
+    def transform(self, X):
+        """Compute the cosine similarity for all pairs of elements in `X`.
+
+        Rows i in `X` are assumed to represent pairs, where
+        `X[i, :n_features]` and `X[i, n_features:]` correspond to their two
+        individual elements. Calling `transform` computes the cosine similarity
+        between between these elements, i.e. that `X_[i]` is the cosine of
```
etzemis: 'between' is duplicated

beard/similarity/pairs.py
```diff
@@ -0,0 +1,212 @@
...
+        """
+        return self
+
+    def transform(self, X):
+        """Compute the absolute difference for all pairs of elements in `X`.
+
+        Rows i in `X` are assumed to represent pairs, where
+        `X[i, :n_features]` and `X[i, n_features:]` correspond to their two
+        individual elements. Calling `transform` computes the absolute
+        difference between between these elements, i.e. that `X_[i]` is the
```
etzemis: the same

beard/similarity/pairs.py
```diff
@@ -0,0 +1,212 @@
...
+        ----------
+        :param element_transformer: transformer
+            The transformer to apply on each element.
+        """
+        self.element_transformer = element_transformer
+
+    def fit(self, X, y=None):
+        """Fit the given transformer on all individual elements in `X`.
+
+        Rows i in `X` are assumed to represent pairs, where
```
MSusik: I think doubled backquotes should be used for code samples. It is worth checking whole Beard code for such misusages.

http://sphinx-doc.org/rest.html

beard/utils/strings.py
```diff
@@ -13,10 +13,13 @@
 
 """
 
-import unicodedata
+import re
 import sys
+import unicodedata
 
 IS_PYTHON_3 = sys.version_info[0] == 3
```
MSusik: I must have missed it in the previous review, but why not simply use `six.u()` instead of this trick?

beard/utils/tests/test_strings.py
```diff
@@ -7,17 +7,38 @@
...
     assert asciify("schröder") == "schroder"
+
+
+def test_normalize_personal_name():
+    """Test of normalize_personal_name"""
+    assert normalize_personal_name("Doe, John") == "doe john"
+    assert normalize_personal_name("Doe, J.") == "doe j"
+    assert normalize_personal_name("Doe, J") == "doe j"
+    assert normalize_personal_name("Doe-Foe, Willem") == "doefoe willem"
+    assert normalize_personal_name("Doe-Foe Willem") == "doe foe willem"
```
MSusik: Why do you need difference in behaviour between the above two examples?

beard/utils/strings.py
glouppe: It does not seem to work -- tests dont pass anymore using this. Maybe we can have a look outside this PR. I agree this is an ugly hack I would rather have an clean solution for this.

beard/utils/tests/test_strings.py
glouppe: In the first case, "Doe-Foe" is the last name, while in the second case, because the comma is missing, we dont know which word is the last name. 

Then the purpose was to transform last names so that "Doe-Foe" and "DoeFoe" are matched together. This seems to happen often for Asian names, where have either a concatenated name or a name formed by multiple words, separated with dashes. 

At this point, this is merely a heuristic anyway, and we should come up something more robust .

beard/utils/strings.py
MSusik: Fine, I will create a PR which changes it.

MSusik:
:+1: from me.

glouppe:
Thanks for the review @MSusik ! Was is it fine for your as well @etzemis ?

kaplun:
Note, as you can imagine lots of work went through the years into normalizing names within Invenio/INSPIRE, have you also considered the existing algorithms as source of inspirations?
I am privately sharing a Google doc with the analysis done so far, in case it can be useful.

MSusik:
@kaplun Thanks! It's definitely worth investigating.

This is the file which contains most relevant work:
https://github.com/inspirehep/invenio/blob/prod/modules/bibauthorid/lib/bibauthorid_name_utils.py

glouppe:
> Note, as you can imagine lots of work went through the years into normalizing names within Invenio/INSPIRE, have you also considered the existing algorithms as source of inspirations?
I am privately sharing a Google doc with the analysis done so far, in case it can be useful.

Thanks, this might be helpful indeed! Let us continue this discussion on #20  however.

etzemis:
:+1: from me too.

PR #22, opened 2015-02-06
Advanced author disambiguation 

### LINKED ISSUE #15: blocking: add verbose mode
It would be nice to add a `verbose` parameter in `BlockClustering`.

glouppe:
* [x] Simplify the basic author disambiguation example
* [x] Advanced author disambiguation, using a distance function learned on paired data
* [x] Add verbose mode in BlockClustering (closes #15)


examples/author_disambiguation_advanced.py
```diff
@@ -0,0 +1,366 @@
...
+    train, test = train_test_split(np.arange(len(X)),
+                                   test_size=0.9, random_state=42)
+    y = -np.ones(len(X), dtype=np.int)
+    y[train] = truth[train]
+
+    clusterer = BlockClustering(
+        blocking=blocking,
+        base_estimator=ScipyHierarchicalClustering(
+            threshold=0.9995,
+            affinity=partial(affinity, estimator=distance_estimator),
```
glouppe: This seems to be a bottleneck, as the learned distance estimator needs to be pickled for every job launched by the clusterer. A workaround is put the estimator variable in global, hence avoiding the transfer. I'll investigate whether there is a cleaner workaround than this ugly hack...

glouppe:
There are still a few things to fix, but this is ready for reviews. @MSusik @etzemis @natsheh

glouppe:
So we have a bottleneck on the coauthors features. Because of the number of authors in physics papers, computing the TF-IDF vectors for the coauthor field is really taking some time. 

By investigating some more, the true culprit is that the current implementation (in `affinity`) is recomputing the vector representation of a signature for every pair in which it appears. We are therefore building O(N^2) such vectors, instead of only O(N) (where N is the number of signatures). I'll try to come up with something to avoid this, while keeping the current API and workflow.

glouppe:
I fixed the bottleneck by transforming only unique elements in `PairTransformer`. 

Ready for reviews.

MSusik:
I would add some comments in `_flatten` function. It is not clear, what the function is doing without a deeper look into it.

Issue #21, opened 2015-02-06
metrics: implement B-CUBED precision/recall/F-score


glouppe:
See http://nlp.uned.es/docs/amigo2007a.pdf section 4.5 for a description.

PR #23, opened 2015-02-20
metrics: b^3 precision / recall / f-score

glouppe:
Implementation of B^3 precision, recall and F-score metrics. These variants have the advantage of not being sensitive to the quadratic effect of cluster size (in contrast with with paired variants, which suffer from this).

* [x] Implementation
* [x] Tests
* [x] Documentation

This addresses #21.

### DIFF
--- a/beard/metrics/__init__.py
+++ b/beard/metrics/__init__.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 #
 # This file is part of Beard.
-# Copyright (C) 2014 CERN.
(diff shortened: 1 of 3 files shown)


glouppe:
Code is ready for review. CC: @natsheh @MSusik

glouppe:
(I also did some cleaning in the `metrics/test/test_clustering.py`)

etzemis:
It is fine by me.

natsheh:
I found it all ok.

glouppe:
Thanks for the reviews. Merging.

PR #24, opened 2015-02-25
WIP Jaccard similarity transformer for paired data

natsheh:
I just add the class for this initial pull


beard/similarity/pairs.py
```diff
@@ -10,6 +10,7 @@
 """Transformers for paired data.
 
 .. codeauthor:: Gilles Louppe <g.louppe@cern.ch>
+.. code co-author:: Hussein AL-NATSHEH <hussein.al.natseh@cern.ch>
```
glouppe: it should be `codeauthor`

beard/similarity/pairs.py
```diff
@@ -273,3 +274,40 @@ def transform(self, X):
         X2 = X[:, n_features:]
 
         return np.abs(X1 - X2)
+
+class SetJaccard(TransformerMixin):
```
glouppe: You should inherit from `BaseEstimator` as well.

beard/similarity/pairs.py
```diff
@@ -273,3 +274,40 @@ def transform(self, X):
...
+	"""Measure How much the 2 elements of a pair have in common"""
```
glouppe: Can you rephrase to be more precise about the mathematical formulation?

beard/similarity/pairs.py
```diff
@@ -273,3 +274,40 @@ def transform(self, X):
...
+        ----------
+        :param X: array-like, shape (n_samples, n_features)
+            Input data.
+
+        Returns
+        -------
+        :returns: Xt which is an array of similarity values
+        """  
+
+        Xt= map(lambda x: float(len(x[0].intersection(x[1]))) / float(len(x[0].union(x[1]))) , X)
```
glouppe: Can you add `from __future__ import division` and remove the casting to `float`?

glouppe: Also, you function assumes that X is in fact `(n_samples, 2)` and that the two values are sets. I would rather assume that you have a `(n_samples, n_features)` array, where `X[i, :n_features/2]` is an array of number where all non-zero features encode for members of the set (e.g., like the representation yield by `CountVectorizer`).

glouppe:
In addition to my comments, can you provide tests for the added functionality?

beard/similarity/pairs.py
glouppe: Can you also rename the class to `JaccardSimilarity`? For consistency with the other transformers defined in the same file.

beard/similarity/pairs.py
```diff
@@ -10,8 +10,12 @@
 """Transformers for paired data.
 
 .. codeauthor:: Gilles Louppe <g.louppe@cern.ch>
+.. codeauthor:: Hussein AL-NATSHEH <h.natsheh@ciapple.com>
 
 """
+from __future__ import division
+from __future__ import print_function
```
glouppe: Can you remove this, as we dont need it?

beard/similarity/pairs.py
```diff
@@ -273,3 +278,66 @@ def transform(self, X):
...
+        Returns
+        -------
+        :returns: self
+        """
+        return self 
+    
+    def transform(self, X):    
+	"""Get the intersection and the union of the 2 feature sets and compute the Jaccard similarity as intersection over union
+         Here, since we have vectorized representation of the input, we calculate the intersection as the dot product
+         and the union as the count of non_zeros of the addition of the 2 vectors
```
glouppe: Please follow PEP8 and PEP257 standards.

beard/similarity/pairs.py
```diff
@@ -273,3 +278,66 @@ def transform(self, X):
...
+            numerator = np.asarray(X1.multiply(X2).sum(axis=1)).ravel()
+            for i in range(n_samples):
+                Xt.append(numerator[i]/(X1[i]+X2[i]).nnz)
+
+        else:
+              for i in range(n_samples):
+                  if np.count_nonzero(X1[i]+ X2[i]) == 0:
+                      Xt.append(0.0)
+                  else:
+                      Xt.append(np.dot(X1[i],X2[i]) / np.count_nonzero(X1[i]+ X2[i]))
```
glouppe: I dont think you need the for loop in either case.

glouppe:
Also, be sure Travis is happy. You can see the log at https://travis-ci.org/inveniosoftware/beard/jobs/52871445

This is currently failing because of formatting errors.

glouppe:
There are several pep257 issues. Can you fix them?
See https://travis-ci.org/inveniosoftware/beard/jobs/53206730#L567

beard/similarity/pairs.py
```diff
@@ -273,3 +274,48 @@ def transform(self, X):
         X2 = X[:, n_features:]
 
         return np.abs(X1 - X2)
+
+
+class JaccardSimilarity(BaseEstimator, TransformerMixin):
+
+    """Similarity as the intersection between 2 sets over the union of them."""
```
glouppe: Can you rephrase as 

```
Jaccard similarity on paired data.

The Jaccard similarity of two elements in a pair is defined as the ratio between the 
size of their intersection and the size of their union.
```

beard/similarity/pairs.py
```diff
@@ -273,3 +274,48 @@ def transform(self, X):
...
+
+    def fit(self, X, y=None):
+        """(Do nothing).
+
+        Parameters
+        ----------
+        :param X: array-like, shape (n_samples, n_features) Input data.
```
glouppe: Missing line break

beard/similarity/pairs.py
```diff
@@ -273,3 +274,48 @@ def transform(self, X):
...
+        return self
+
+    def transform(self, X):
+        """Compute the intersection over the union.
+
+        Get the intersection and the union of the 2 feature sets
+        and compute the Jaccard similarity as intersection over union
+        Here, since we have vectorized representation of the input, we
+        calculate the intersection as the dot product and the union as
+        the count of non_zeros of the addition of the 2 vectors.
```
glouppe: Can you rephrase as 

```
Compute the Jaccard similarity for all pairs of elements in ``X``.

Rows i in ``X`` are assumed to represent pairs, where
``X[i, :n_features]`` and ``X[i, n_features:]`` correspond to their two
individual elements, each representing a set. Calling ``transform`` computes the Jaccard
similarity between these sets, i.e. such that ``Xt[i]`` is the Jaccard similarity of
``X[i, :n_features]`` and ``X[i, n_features:]``.
```

beard/similarity/pairs.py
```diff
@@ -273,3 +274,48 @@ def transform(self, X):
...
+
+        Parameters
+        ----------
+        :param X: array-like, shape (n_samples, n_features)
+            Input data.
+
+        Returns
+        -------
+        :returns: Xt which is an array of similarity values
```
glouppe: Rephrase as

```
        :returns Xt: array-like, shape (n_samples, 2 * n_features_prime)
            The transformed data.
```

beard/similarity/pairs.py
```diff
@@ -273,3 +274,48 @@ def transform(self, X):
...
+        """
+        X = sp.csr_matrix(X)
```
glouppe: If `X` is a dense array, could you avoid making that copy?

beard/similarity/pairs.py
```diff
@@ -273,3 +274,48 @@ def transform(self, X):
...
+        n_samples, n_features_all = X.shape
+        n_features = n_features_all // 2
+        X1 = X[:, :n_features]
+        X2 = X[:, n_features:]
+        numerator = np.asarray(X1.multiply(X2).sum(axis=1)).ravel()
+        denominator = (X1 + X2).getnnz(axis=1)
```
glouppe: `getnnz` is not reliable in this context, as it counts explicitely encoded zero values.

glouppe: Also, what happens when all values are zeros?

beard/similarity/pairs.py
MSusik: BTW, those reversed apostrophes are `reStructuredText` syntax. 
http://sphinx-doc.org/rest.html

glouppe:
Once you have adressed the comments, can you squash all the commits?

beard/similarity/pairs.py
natsheh: if all values are zeros in the numerator, the result would be zeros for all. If the denominator has any zero value, the corresponding pair similarity would be zero.

glouppe: Wouldnt it crash if you divide by zero? Please add a test for that to ensure the proper behaviour.

beard/similarity/pairs.py
```diff
@@ -273,3 +274,54 @@ def transform(self, X):
...
+        ``X[i, n_features:]``.
+
+        Parameters
+        ----------
+        :param X: array-like, shape (n_samples, n_features)
+            Input data.
+
+        Returns
+        -------
+        :returns: Xt array-like, shape (n_samples, 1) The transformed data.
```
glouppe: Add a line break before "The transformed data."

beard/similarity/pairs.py
```diff
@@ -273,3 +274,75 @@ def transform(self, X):
...
+            X2 = X[:, n_features:]
+            numerator = np.asarray(X1.multiply(X2).sum(axis=1)).ravel()
+            X_sum = X1 + X2
+            X_sum.data[X_sum.data != 0.] = 1
+            M = X_sum.sum(axis=1)
+            A = M.getA()
+            denominator = A.reshape(-1,)
+            Xt = numerator / denominator
+            return np.array(Xt).reshape(-1, 1)
+        X = np.array(X)  # Just to make sure that X array is a numpy array
```
glouppe: Please move this and the following lines within an `else:` clause to make it clearer that you are dealing with the dense case.

beard/similarity/tests/test_pairs.py
```diff
@@ -87,3 +88,14 @@ def test_absolute_difference():
 
     Xt = AbsoluteDifference().fit_transform(sp.csr_matrix(X))
     assert_array_almost_equal(Xt, [[0, 0], [1, 1], [0, 0], [1, 1]])
+
+
+def test_JaccardSimilarity():
+    X = np.array([[0, 1, 0, 1, 0, 1, 0, 1],
+                  [0, 0, 1, 1, 0, 1, 0, 1],
+                  [0, 1, 0, 1, 0, 0, 1, 0],
+                  [1, 0, 1, 1, 1, 1, 0, 1]])
```
glouppe: Please add a test with a row full of zeros.

beard/similarity/pairs.py
```diff
@@ -273,3 +274,75 @@ def transform(self, X):
...
+        if sparse and not sp.isspmatrix_csr(X):
+            X = X.tocsr()
+        if sparse:
+            if X.data[X.data != 0.].sum() == 0:
+                return np.zeros(n_samples).reshape(-1, 1)
+            X1 = X[:, :n_features]
+            X2 = X[:, n_features:]
+            numerator = np.asarray(X1.multiply(X2).sum(axis=1)).ravel()
+            X_sum = X1 + X2
+            X_sum.data[X_sum.data != 0.] = 1
```
MSusik: I would suggest to add some empty lines in between. It's hard to read such a blob.

beard/similarity/pairs.py
glouppe: Some could certainly also be factored out, there are common steps between sparse and dense.

glouppe:
By the way, what would think about using `sklearn.metrics.jaccard_similarity_score` simply over `X1` and `X2`?

PR #25, opened 2015-02-27
similarity: fuzzy similarity

etzemis:


beard/similarity/fuzzy.py
```diff
@@ -0,0 +1,107 @@
...
+"""
+
+import numpy as np
+import scipy.sparse as sp
+import operator
+
+from sklearn.base import TransformerMixin
+
+
+class FuzzySimilarity(TransformerMixin):
```
glouppe: Can you put this class in `pairs.py`, as it concerns paired data only?

beard/similarity/fuzzy.py
```diff
@@ -0,0 +1,107 @@
...
+            X = X.todense()
+
+        X1 = X[:, :n_features]
+        X2 = X[:, n_features:]
+
+        summ = 0
+        # for each document in X1
+        for row, u in enumerate(X1):
+            # for each word in document u
+            for pos, w in enumerate(u):
```
glouppe: You can use `nonzero` to iterate over defined values only. And this would work with sparse matrices too.

beard/similarity/fuzzy.py
```diff
@@ -0,0 +1,107 @@
...
+                if w > 0.0:
+                    w_word = self.vocabulary[pos]
+                    min_d, n_index = self._find_distance(w_word, X2[row, :])
+                    summ += w * X2[row, n_index] * min_d
```
glouppe: I am not sure to understand. So you are basically computing a weighted dot product. I think it would be better it this was normalized, to have something close to `CosineSimilarity`.

beard/similarity/fuzzy.py
```diff
@@ -0,0 +1,107 @@
...
+        return summ
+
+    def _find_distance(self, u, words):
+        """Find the distance between u and words according to threshold.
```
glouppe: I think you should rather say that you are finding the closest word.

glouppe:
In general, you should make sure your code works for sparse matrices. Input vectors will be yielded by TFIDF or count in the most general case, and these vectors are sparse.

glouppe:
How is it going @etzemis, anything blocking you?

glouppe:
Can you push your latest code, even if it is not yet fully ready, so that we can start reviewing?

beard/similarity/pairs.py
```diff
@@ -273,3 +276,124 @@ def transform(self, X):
...
+        Parameters
+        ----------
+        :param vocabulary:
+            From word to feature_id or opposite.
+
+        :param similarity:
+            Similarity function which returns a number between 0 and 1.
+
+        :param threshold:
+            Threshold parameter that belongs to R.
```
glouppe: What is R?

glouppe: Ha, I assume you mean R=real numbers.

In that case, just simply document the parameter as follows

```
:param threshold: float
    ...
```

Please also describe what is this value thresholding.

beard/similarity/pairs.py
```diff
@@ -273,3 +276,124 @@ def transform(self, X):
         X2 = X[:, n_features:]
 
         return np.abs(X1 - X2)
+
+
+class FuzzySimilarity(TransformerMixin):
+
+    """Implements Fuzzy Similarity (Soft TF-IDF) metric."""
```
glouppe: Add a short description of the algorithm. It is not clear otherwise what is this class exactly doing.

beard/similarity/pairs.py
```diff
@@ -273,3 +276,124 @@ def transform(self, X):
...
+        sparse = sp.issparse(X)
+
+        if sparse and not sp.isspmatrix_csr(X):
+            X = X.tocsr()
+
+        X1 = X[:, :n_features]
+        X2 = X[:, n_features:]
+
+        X1_closest = np.zeros((n_samples, n_features))
+        Similarities = np.zeros((n_samples, n_features))
```
glouppe: `similarities`

beard/similarity/pairs.py
```diff
@@ -273,3 +276,124 @@ def transform(self, X):
...
+            Xt[denominator == 0.0] = 0.0
+
+        return Xt.reshape((n_samples, 1))
+
+        
+
+    def _find_closest_neighbour(self, u, words):
+        """Find the word in words with the maximized similarity with u.
+
+        If threshold is 'None', return the maximum similarity between u
```
glouppe: Can you make `threshold` a function parameter instead of getting it from `self`?  (in fact `_find_closest_neighbour` does not need to be a method)

glouppe: the same for `vocabulary` and `similarity`.

beard/similarity/tests/test_pairs.py
```diff
@@ -87,3 +93,201 @@ def test_absolute_difference():
...
+    arr = tfidf.toarray()
+    X = np.array([[item for sublist in arr for item in sublist]])
+
+    Y = FuzzySimilarity(vocabulary, sim_f, thrs).fit_transform(X)
+    Y_prime = CosineSimilarity().fit_transform(X)
+
+    assert_array_almost_equal(Y, Y_prime)
+
+
+def _simple_similarity_function(w1, w2):
```
glouppe: Define this function inside the test above

glouppe:
I see that sparse support is not yet implemented. Ping me if you need help.

PR #26, opened 2015-03-09
metrics: bug fix in jaro metric

petros-ioannidis:
Wrong computation of words that contain double letters in a sequence.

Signed-off-by: Petros Ioannidis <petros.ioannidis91@gmail.com>


glouppe:
Thanks for the patch. Merging.

PR #27, opened 2015-03-10
examples: operationalization of the author disambiguation example

glouppe:
* [x] Add command line arguments for the advanced disambiguation example.
* [x] Switch to JSON format for input/output (rather than pickles), allowing for an easier use in external contexts.
* [x] Documentation.

### DIFF
--- a/examples/author_disambiguation_advanced.py
+++ b/examples/author_disambiguation_advanced.py
@@ -427,7 +427,8 @@ if __name__ == "__main__":
 
         indices = {}
         X = np.empty((len(signatures), 1), dtype=np.object)
-        for i, signature in enumerate(signatures.values()):
+        for i, signature in enumerate(sorted(signatures.values(),
+                                             key=lambda s: s["signature_id"])):
             X[i, 0] = signature
             indices[signature["signature_id"]] = i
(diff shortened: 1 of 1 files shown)


glouppe:
Okay, so the example is now properly working using JSON input/output format. I will document everything tomorrow so that you @natsheh and @MSusik can move on with your own analyses.

glouppe:
Documentation is now complete. See https://github.com/glouppe/beard/blob/author-disambiguation-ops/examples/author_disambiguation_advanced.py#L55 for a description of the expected input format.

glouppe:
Reviews are welcome :) @natsheh @MSusik @etzemis

MSusik:
Clearer than previous version, :+1:

glouppe:
Thanks for the review! @natsheh would you have time today for a second review? thanks

natsheh:
Looks great! I just have not a critical suggestion:
it may be more interesting to use "sklearn.externals.joblib"’s replacement of pickle (joblib.dump & joblib.load), which is more efficient on big data. I have noticed that on Windows enviroment, I cannot load the pickle dump file.

PR #31, opened 2015-03-27
Adding unsupervised capability to HAC thresholding by maximizing  Silhouette score

natsheh:


glouppe:
Can you rebase on master to avoid including my commits?

PR #32, opened 2015-03-30
utils: name tokenization

MSusik:
* Provides a tokenizing function for future blocking algorithm.

### DIFF
--- a/beard/utils/names.py
+++ b/beard/utils/names.py
@@ -10,10 +10,14 @@
 """Helper functions for handling personal names.
 
 .. codeauthor:: Gilles Louppe <g.louppe@cern.ch>
+.. codeauthor:: Mateusz Susik <mateusz.susik@cern.ch>
 
 """
 
+import functools
 import re
+
+from ..ext.metaphone import dm
 
 from .misc import memoize
 from .strings import asciify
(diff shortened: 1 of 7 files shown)


MSusik:
The directory `ext` contains the only implementation of Double Metaphore algorithm working for both Python2 and Python3. After the working implementation is available in, for example, `fuzzy` package, we should switch to it.

I made `pep` omit this directory.

The algorithm will come in next PR.

beard/utils/names.py
```diff
@@ -46,3 +50,69 @@ def normalize_name(name):
...
+    # Handle the "miagkii znak" in russian names.
+    matches = re.findall(r"^([^',]*)'([a-z].*)", name)
+    if matches:
+        name = matches[0][0] + matches[0][1]
+
+    # Remove apostrophes
+    name = RE_APOSTROPHES.sub(' ', name)
+
+    # Extract surname and name
+    tokens = name.split(',')
```
glouppe: What if we are given a single name, without commas in the string?

glouppe: Your function should work, but this makes me think we should maybe update the docstring to describe the input format of names.

beard/utils/names.py
```diff
@@ -46,3 +50,69 @@ def normalize_name(name):
...
+                    'bath', 'bat', 'ben', 'bet', 'bin', 'bint', 'd', 'da',
+                    'dall', 'dalla', 'das', 'de', 'degli', 'del', 'dell',
+                    'della', 'dem', 'den', 'der', 'di', 'do', 'dos', 'du', 'e',
+                    'el', 'i', 'ibn', 'im', 'jr', 'l', 'la', 'las', 'le',
+                    'los', 'm', 'mac', 'mc', 'mhic', 'mic', 'o', 'ter', 'und',
+                    'v', 'van', 'vom', 'von', 'zu', 'zum', 'zur'}
+
+
+@memoize
+def tokenize_name(name):
```
glouppe: I think we should be more explicit in the function name. Something like `dm_tokenize_name` maybe.

glouppe:
Besides the small comments regarding documentation, this looks good to me. +1 for merge

beard/utils/names.py
```diff
@@ -46,3 +50,69 @@ def normalize_name(name):
...
+
+RE_APOSTROPHES = re.compile('\'+')
+RE_REMOVE_NON_CHARACTERS = re.compile('[^a-zA-Z\',\s]+')
+DROPPED_PREFIXES = {'a', 'ab', 'am', 'ap', 'abu', 'al', 'auf', 'aus', 'bar',
+                    'bath', 'bat', 'ben', 'bet', 'bin', 'bint', 'd', 'da',
+                    'dall', 'dalla', 'das', 'de', 'degli', 'del', 'dell',
+                    'della', 'dem', 'den', 'der', 'di', 'do', 'dos', 'du', 'e',
+                    'el', 'i', 'ibn', 'im', 'jr', 'l', 'la', 'las', 'le',
+                    'los', 'm', 'mac', 'mc', 'mhic', 'mic', 'o', 'ter', 'und',
+                    'v', 'van', 'vom', 'von', 'zu', 'zum', 'zur'}
```
glouppe: When you filter on a,d,e,l,o,v, how do you know you are not removing a name initial?

MSusik: The affixes are dropped only in the surname.

beard/utils/tests/test_names.py
```diff
@@ -30,6 +33,51 @@ def test_normalize_name():
...
+    """Test correct removal of the common affixes."""
+    assert dm_tokenize_name("von und zu Hohenstein, F.") == \
+        dm_tokenize_name("Hohenstein, F.")
+    # If the name consists of only the common prefixes, don't drop it, as
+    # it might actually be the correct surname.
+    assert dm_tokenize_name("Ben, Robert") == ((dm(u"Ben")[0],),
+                                               (dm(u"Robert")[0],))
+    # Don't drop affixes among the first names.
+    assert dm_tokenize_name("Robert, L. W.") == ((dm(u"Robert")[0],),
+                                                 (dm(u"L")[0], dm(u"W")[0]))
```
MSusik: I added a test to cover the case where the first name initial is included in the `DROPPED_AFFIXES` set.

beard/utils/names.py
MSusik: A more detailed description how the string without commas are handled is available below.

glouppe:
Great, thanks for the quick changes!

@natsheh @etzemis Please review :)

beard/utils/names.py
```diff
@@ -46,3 +50,77 @@ def normalize_name(name):
...
+                   'bath', 'bat', 'ben', 'bet', 'bin', 'bint', 'd', 'da',
+                   'dall', 'dalla', 'das', 'de', 'degli', 'del', 'dell',
+                   'della', 'dem', 'den', 'der', 'di', 'do', 'dos', 'du', 'e',
+                   'el', 'i', 'ibn', 'im', 'jr', 'l', 'la', 'las', 'le',
+                   'los', 'm', 'mac', 'mc', 'mhic', 'mic', 'o', 'ter', 'und',
+                   'v', 'van', 'vom', 'von', 'zu', 'zum', 'zur'}
+
+
+@memoize
+def dm_tokenize_name(name):
```
natsheh: I suggest having a tokenize_name function that could be reused without the Metaphone step. Other steps like handeling apotrophes and prefixes should be parametrized. Additional parametrized step could also be added which is spliting on the caps before lower-casing the token. So that 'MacDonald' example can be treated and could also pass throught eh prefixes step. So the user can deside on this parametrized steps using a scoring metric over her/his own database.

MSusik: Done, any other suggestions?

beard/utils/names.py
```diff
@@ -46,3 +50,103 @@ def normalize_name(name):
...
+def tokenize_name(name, handle_soft_sign=True, drop_common_affixes=True):
+    """Preprocess the name and create tokens from the name.
+
+     Parameters
+    ----------
+    :param name: string
+        Name of the author. Usually it should be in the format:
+        surnames, first names.
+    :param handle_soft_sign: boolean
+        Should the case of cyryllic soft sign be handled.
```
natsheh: Isn't it 'cyrillic' instead of 'cyryllic'

beard/utils/names.py
```diff
@@ -46,3 +50,103 @@ def normalize_name(name):
...
+    # Use double metaphone
+    tokens = tuple(map(lambda x: tuple(map(lambda y: dm(y)[0], x)),
+                   tokens))
+
+    return tokens
+
+
+@memoize
+def tokenize_name(name, handle_soft_sign=True, drop_common_affixes=True):
+    """Preprocess the name and create tokens from the name.
```
glouppe: Can you be more specific about what is done? "preprocess" is too vague

natsheh:
+1 for me

glouppe:
Merging. Thanks @MSusik and @natsheh for review.

PR #34, opened 2015-04-01
Adding silhouette_score for unsupervised thresholding

natsheh:


natsheh:
Hello @glouppe @MSusik @etzemis. Could you please review? Thanks.

glouppe:
Thanks for making it a nice PR :) I'll try checking that tomorrow.

beard/clustering/tests/test_wrappers.py
```diff
@@ -31,6 +34,14 @@ def test_scipy_hierarchical_clustering():
     random_state = check_random_state(42)
     X, y = make_blobs(centers=4, shuffle=False, random_state=random_state)
 
+    # Unsupervised clustering
```
MSusik: This tests: `test_scipy_hierarchical_clustering` and `test_scipy_hierarchical_clustering_semi_supervised` contain a lot of use cases. It might be worth to split it. You can use `pytest.fixture` to get the same `X` and `y` for every test case.

beard/clustering/wrappers.py
```diff
@@ -80,8 +86,17 @@ def __init__(self, method="single", affinity="euclidean", threshold=None,
...
-            (when y!=None).
+            The scoring function to maximize in either (semi-)supervised
+            clustering if y is not None, or unsupervised otherwise (using
+            "silhouette_score as an example when y is None or all y =-1).
+            In case of silhouette_score, the metric param of this function
+            must be set to 'precomputed'.
+            See sklearn.metrics.silhouette_score for further details.
+
+        :param affinity_score: boolean
+            A flag that must be Ture if the scoring function requiers the
```
MSusik: True, requires, otherwise.

beard/clustering/wrappers.py
```diff
@@ -80,8 +86,17 @@ def __init__(self, method="single", affinity="euclidean", threshold=None,
             See scipy.cluster.hierarchy.fcluster for further details.
 
         :param scoring: callable
-            The scoring function to maximize in (semi-)supervised clustering
-            (when y!=None).
+            The scoring function to maximize in either (semi-)supervised
+            clustering if y is not None, or unsupervised otherwise (using
+            "silhouette_score as an example when y is None or all y =-1).
```
MSusik: Using sphinx syntax -> using ``silhouette_score`` as an example when ``y`` is None or all y = -1.

beard/clustering/wrappers.py
```diff
@@ -80,8 +86,17 @@ def __init__(self, method="single", affinity="euclidean", threshold=None,
...
+            In case of silhouette_score, the metric param of this function
```
MSusik: metric? I believe the argument is called ``affinity``. Also, it is a description, so no reason to use `param` instead of `parameter`.

beard/clustering/wrappers.py
```diff
@@ -80,8 +86,17 @@ def __init__(self, method="single", affinity="euclidean", threshold=None,
...
+            must be set to 'precomputed'.
```
MSusik: If you want italics inside RTD, use stars instead of quotes:
http://docutils.sourceforge.net/docs/user/rst/quickref.html#inline-markup

beard/clustering/wrappers.py
```diff
@@ -151,8 +170,28 @@ def fit(self, X, y=None):
                                       criterion=self.criterion,
                                       depth=self.depth, R=self.R,
                                       monocrit=self.monocrit)
+                if ground_truth:
+                    train = (y != -1)
+
+                    if train.sum() == 0:
```
MSusik: Isn't `train.any()` better? Is this line doing anything?

beard/clustering/wrappers.py
```diff
@@ -151,8 +170,28 @@ def fit(self, X, y=None):
...
+                    n_labels = len(np.unique(labels))
+                    n_samples = Xs.shape[0]
 
-                score = self.scoring(y[train], labels[train])
+                    if 1 < n_labels < n_samples:
+                        score = self.scoring(Xs, labels)
+                    else:
+                        score = -np.inf
+                else:
+                    # could have a scoring function that accepts only lables
```
MSusik: Typo: `labels`

beard/clustering/wrappers.py
natsheh: Could you please be more precised about this commit @MSusik

MSusik: The line that I commented (and one below) contain three misspellings.

beard/clustering/wrappers.py
natsheh: I will remove this clause as we already have this check embeded in ground_truth condition.

natsheh:
Hi @MSusik! I have corresponded to your feedbacks. Could you please validate and tell me if there is anything else I can enhance? Thanks.

MSusik:
Looks good to me!

beard/clustering/wrappers.py
```diff
@@ -151,8 +170,26 @@ def fit(self, X, y=None):
...
-
-                score = self.scoring(y[train], labels[train])
+                if ground_truth:
+                    train = (y != -1)
+
+                    if not self.affinity_score:
+                        score = self.scoring(y[train], labels[train])
+                    else:
+                        # For scoring function that accepts affinity
+                        score = self.scoring(y[train], labels[train], X)
```
glouppe: I would put X as first argument, for consistency with the case when `ground_truth is False`

beard/clustering/wrappers.py
```diff
@@ -80,8 +86,17 @@ def __init__(self, method="single", affinity="euclidean", threshold=None,
...
 
         :param scoring: callable
-            The scoring function to maximize in (semi-)supervised clustering
-            (when y!=None).
+            The scoring function to maximize in either (semi-)supervised
+            clustering if y is not None, or unsupervised otherwise (using
+            ``silhouette_score`` as an example when y is None or all y =-1).
+            In case of silhouette_score, 'affinity' parameter of this
+            function must be set to *precomputed*.
+            See sklearn.metrics.silhouette_score for further details.
```
glouppe: We should be explicit about the expected input arguments. 

We should state that the scoring function needs to have the following prototype:
`scoring(labels_true, labels_pred)`, `scoring(labels_pred)`, `scoring(Xs, labels_true, labels_pred)` or  `scoring(Xs, labels_pred)` depending on the case and the value of `affinity_score`.

beard/clustering/wrappers.py
```diff
@@ -151,8 +170,26 @@ def fit(self, X, y=None):
...
+                    n_labels = len(np.unique(labels))
+                    n_samples = Xs.shape[0]
+
+                    if 1 < n_labels < n_samples:
+                        score = self.scoring(Xs, labels)
+                    else:
+                        score = -np.inf
+                else:
+                    # could have a scoring function that accepts only labels
+                    return self
```
glouppe: Indeed, I would handle this by setting `score = self.scoring(labels)` instead of quitting.

glouppe:
Thanks for refactoring the tests. It is much better like this :) 

+1 for merge once my other comments are addressed.

beard/clustering/wrappers.py
natsheh: Actually, I had that before but I replace it with quitting and added a commit above since I got a test error having the default scoring function "b3_f_score()"  taking exactly 2 arguments. If we don't set a default scoring we will also have this error. The test cases I reach this error with is when we specify the number of clusters or the  treshold. I would need to change the logic of handeling these 2 cases in order to make this change. Is it fine to do that?

glouppe: You should keep `b3_f_score` as default scoring, along with `affinity_score=False` by default. It should trigger the correct path, shouldnt it?

glouppe: > The test cases I reach this error with is when we specify the number of clusters or the treshold. I would need to change the logic of handeling these 2 cases in order to make this change. Is it fine to do that?

Yes, as long as the output remains the same and we can keep doing what was allowed before, it is fine with me.

natsheh:
I have made some changes covering your feecback comments. The main changes were to fit with case of not passing any parameters except n_clusters or  treshold. @glouppe can you please double check and approve the PR? Thanks.

glouppe:
Thanks for the changes! Merging :beers:

PR #36, opened 2015-04-07
WIP Finding the best default threshold for HAC

natsheh:


MSusik:
A lot of things in this PR seem to be duplicated from `author_disambiguation_advanced.py`

glouppe:
Indeed, I would try instead to plugin the default threshold strategy into ScipyHierarchicalClustering rather than duplicating the example.

glouppe:
What shall we do with this PR? Shall I close it @natsheh ?

PR #38, opened 2015-04-10
Bug Fix: number of arguments that the scoring function takes

natsheh:


natsheh:
Hi @glouppe @MSusik Here is the fix!

glouppe:
I think a better fix would be to simply to `scoring=None` by default, and then resort `threshold` or `n_clusters`. If on the other hand `scoring is not None`, then we can assume it has the proper prototype depending on the case we deal with.

natsheh:
scoring is None by default. The problem raises when the user uses B3 scoring while all y are -1. It might be not easy for the user to make this validation before she/he decide to use B3 specially in blocking as in the example.

glouppe:
Yeah, but this makes the interface more complicated. Too much magic is not good. I would instead maintain what I said before and then amend the Blocking clusterer to handle this case specifically. What do you think?

natsheh:
Yes. I think it is a better idea to update Blocking clusterer instead. I guess in that case it is better to close this PR.

glouppe:
Or simply amend your commit, to have the discussion all gathered here in this thread only.

beard/clustering/wrappers.py
```diff
@@ -20,6 +20,8 @@
 from sklearn.base import BaseEstimator
 from sklearn.base import ClusterMixin
 
+import inspect
```
glouppe: This is not used and can be removed

beard/clustering/wrappers.py
```diff
@@ -79,17 +81,22 @@ def __init__(self, method="single", affinity="euclidean",
...
-            - ground_truth but not affinity: scoring(labels_true, labels_pred)
-            - affinity but not ground_truth: scoring(X, labels_pred)
-            - none: scoring(labels_pred).
-
-        :param affinity_score: boolean
-            A flag that must be True if the scoring function requires the
-            affinity as an input. False otherwise.
+            threshold. There are 6 possible cases based on ground_truth
+            availability and the scoring_data parameter:
+            - ground_truth and row: scoring(Xr, label_true, label_pred);
```
glouppe: It is not _row_ but _raw_

beard/clustering/wrappers.py
```diff
@@ -79,17 +81,22 @@ def __init__(self, method="single", affinity="euclidean",
...
-            threshold. There are 4 possibles cases based on data availability:
-            - ground_truth and affinity: scoring(X, labels_true, labels_pred)
-            - ground_truth but not affinity: scoring(labels_true, labels_pred)
-            - affinity but not ground_truth: scoring(X, labels_pred)
-            - none: scoring(labels_pred).
-
-        :param affinity_score: boolean
-            A flag that must be True if the scoring function requires the
-            affinity as an input. False otherwise.
+            threshold. There are 6 possible cases based on ground_truth
```
glouppe: based on the availability of ground truth data passed in `y`

beard/clustering/wrappers.py
glouppe: What is Xr and Xa? please define all notations

beard/clustering/wrappers.py
```diff
@@ -121,25 +128,29 @@ def fit(self, X, y=None):
...
             X = X[i, j]
             self.linkage_ = hac.linkage(X, method=self.method)
-        else:
+
+        elif self.scoring_data != "affinity":
             self.linkage_ = hac.linkage(X,
                                         method=self.method,
                                         metric=self.affinity)
+        else:
+            raise ValueError("Clustering error. Check the inputs combination.")
```
glouppe: Please be more informative about what went wrong and how to fix it

beard/clustering/wrappers.py
```diff
@@ -79,17 +81,22 @@ def __init__(self, method="single", affinity="euclidean",
             The statistics upon which non-singleton i is thresholded.
             See scipy.cluster.hierarchy.fcluster for further details.
 
+         :param scoring_data: string
+            The form of the input data
+            - "row": Array of samples;
```
glouppe: _raw_

glouppe: You should also say that it can be None

beard/clustering/wrappers.py
```diff
@@ -168,31 +179,38 @@ def fit(self, X, y=None):
...
+                        score = scoring(Xr, y[train], labels[train])
+
+                    elif self.scoring_data == "affinity":
+                        score = scoring(Xa, y[train], labels[train])
+
                     else:
-                        score = scoring(Xs, y[train], labels[train])
+                        score = scoring(y[train], labels[train])
+
+                elif 1 < n_labels < n_samples:
```
glouppe: Why do we need this check actually? It seems you are making assumptions about the scoring function by doing that

beard/metrics/clustering.py
```diff
@@ -19,9 +20,23 @@
 from operator import mul
 from itertools import groupby
 
+from sklearn.metrics import silhouette_score
 from sklearn.metrics.cluster.supervised import check_clusterings
 
 
+def silhouette(X, labels, metric="precomputed"):
```
glouppe: Please keep the `_score` suffix. And to avoid name clashes, import the function from sklearn using
`from sklearn.metrics import silhouette_score as sklearn_silhouette_score`

beard/metrics/clustering.py
```diff
@@ -19,9 +20,23 @@
...
+    """Compute silhouette_score with an extension of number of lables condition.
```
glouppe: Please specify what are these conditions

beard/metrics/clustering.py
```diff
@@ -19,9 +20,23 @@
...
+
+    Please check sklearn.metrics.silhouette_score for the documentation
+    """
+    n_samples = X.shape[0]
+    n_labels = len(np.unique(labels))
```
glouppe: please rename to `n_clusters`

natsheh:
Done! @glouppe

glouppe:
You should add a test checking all 6 cases.

beard/clustering/tests/test_wrappers.py
```diff
@@ -149,7 +162,7 @@ def test_sp_hierarchical_clustering_semi_supervised_no_label(semiSupervised):
     X, y = semiSupervised
 
     # All labels are unknown, hence it should yield a single cluster
-    clusterer = ScipyHierarchicalClustering()
+    clusterer = ScipyHierarchicalClustering(scoring_data="row")
```
glouppe: `row` is not valid

beard/clustering/tests/test_wrappers.py
```diff
@@ -27,12 +27,25 @@
...
 from beard.metrics import b3_f_score
 
-from sklearn.metrics import silhouette_score
+from beard.metrics import silhouette_score
 from functools import partial
 
 
-# Test wrapper of Scipy hierarchical clustering
+# Defining some scoring functions for related test cases
+def pred_labels_score(pred_labels):
```
glouppe: Can you make this private and define it close to its usage?

beard/clustering/tests/test_wrappers.py
```diff
@@ -27,12 +27,25 @@
...
+    """Scoring function that takes 1 exact argument: labels."""
+    return -np.inf
+
+
+def groundTruth_X_score(X, true_labels, pred_labels):
```
glouppe: same here

beard/clustering/tests/test_wrappers.py
```diff
@@ -27,12 +27,25 @@
...
+    """Scoring function that takes 3 exact arguments."""
+    # For the sake of the case test, X is ignored. So, it will be
+    # as using 2 exact arguments function like b3_f_score.
```
MSusik: `as` vs `as if`?
`2 exact arguments function` vs `function that takes exactly 2 arguments`

beard/clustering/tests/test_wrappers.py
```diff
@@ -27,12 +27,11 @@
...
 
-from sklearn.metrics import silhouette_score
+from beard.metrics import silhouette_score
 from functools import partial
 
 
 # Test wrapper of Scipy hierarchical clustering
-
 @pytest.fixture
 def testData():
```
glouppe: Can you avoid using camelCase? Prefer `test_data`

beard/clustering/tests/test_wrappers.py
```diff
@@ -41,20 +40,124 @@ def testData():
...
+def unsupervised():
+    """Preparing a test data."""
+    random_state = check_random_state(42)
+
+    X, y = make_blobs(centers=4, shuffle=False, random_state=random_state)
+    return X
+
+
+@pytest.fixture
+def semiSupervised():
```
glouppe: same here

beard/clustering/tests/test_wrappers.py
```diff
@@ -41,20 +40,124 @@ def testData():
...
+    """Preparing a test data for semi_Supervised cases."""
```
glouppe: semi-supervised

beard/clustering/tests/test_wrappers.py
```diff
@@ -41,20 +40,124 @@ def testData():
...
+    X, y = make_blobs(centers=4, shuffle=False, random_state=random_state)
+    mask = random_state.randint(2, size=len(y)).astype(np.bool)
+    y[mask] = -1
+    d = euclidean_distances(X)
+    d = (d + d.T) / 2.0
+    d /= d.max()
+    return d, y
+
+
+# Defining a scoring function for the following 2 test cases
```
glouppe: I dont understand what are the two cases

glouppe: You can remove this

beard/clustering/tests/test_wrappers.py
```diff
@@ -41,20 +40,124 @@ def testData():
...
+def _groundTruth_X_score(X, true_labels, pred_labels):
```
glouppe: Please following the naming conventions: `labels_true` and `labels_pred`

beard/clustering/tests/test_wrappers.py
```diff
@@ -41,20 +40,124 @@ def testData():
...
+    """Scoring function that takes exactly 3 arguments."""
+    # For the sake of the test cases, X will be ignored. So, it will be like
+    # using a function that takes exactly 2 arguments.
+    score = b3_f_score(true_labels, pred_labels)
+    return score
+
+
+def test_scipy_HAC_semi_supervised_labels_affinity(semiSupervised):
```
glouppe: Can you name all tests in a consistent way?

beard/clustering/tests/test_wrappers.py
```diff
@@ -41,20 +40,124 @@ def testData():
...
+    """Test semi-supervised learning for sp.HAC when scoring_data=affinity."""
+    X, y = semiSupervised
+
+    # We should find all 4 clusters
+    clusterer = ScipyHierarchicalClustering(scoring=_groundTruth_X_score,
+                                            scoring_data="affinity",
+                                            affinity=euclidean_distances)
+    clusterer.fit(X, y)
+    labels = clusterer.labels_
+    assert_array_equal([25, 25, 25, 25], np.bincount(labels))
```
glouppe: I am sorry to delay this further, but these tests are not really testing anything. You should make sure that the proper data is passed to the scoring function. E.g,:

```
def my_test():
    def scoring(Xa, labels_true, labels_pred):
         assert Xa.shape[0] == Xa.shape[1]  # Check that it is an affinity matrix
         return b3_f_score(labels_true, labels_pred)

    X, y = semiSupervised

    # We should find all 4 clusters
    clusterer = ScipyHierarchicalClustering(scoring=scoring,
                                            scoring_data="affinity",
                                            affinity=euclidean_distances)
    clusterer.fit(X, y)
    labels = clusterer.labels_
    assert_array_equal([25, 25, 25, 25], np.bincount(labels))
```

beard/clustering/tests/test_wrappers.py
```diff
@@ -41,20 +40,124 @@ def testData():
...
+
+
+def test_scipy_HAC_semi_supervised_labels_raw(semiSupervised):
```
glouppe: The same here

beard/clustering/wrappers.py
```diff
@@ -79,17 +79,24 @@ def __init__(self, method="single", affinity="euclidean",
...
-            - none: scoring(labels_pred).
-
-        :param affinity_score: boolean
-            A flag that must be True if the scoring function requires the
-            affinity as an input. False otherwise.
+            threshold. There are 6 possible cases based on the availability of
+            ground truth data passed in y and the scoring_data parameter value:
+            (Notations definition: Xr means X when it contains the raw samples;
+                Xa means X when it is the affinity matrix of the samples.)
+            - ground_truth and Xr: scoring(Xr, label_true, label_pred);
```
glouppe: Rephrase it as follows:

- Ground truth data and `scoring_data == "raw"`: 
- No ground truth data and `scoring_data == "raw"`: 
- Ground truth data and `scoring_data == "affinity"`: 
- No ground truth data and `scoring_data == "affinity"`: 
- Ground truth data and `scoring_data is None`: 
- No ground truth data and `scoring_data is None`:

MSusik:
+1

beard/clustering/tests/test_wrappers.py
MSusik: Python naming convenions don't allow for names like `semiSupervised` (mixed case)

https://www.python.org/dev/peps/pep-0008/#prescriptive-naming-conventions

Rephrase it to `semi_supervised` or `semisupervised` everywhere where it appears.

beard/clustering/wrappers.py
```diff
@@ -79,17 +79,32 @@ def __init__(self, method="single", affinity="euclidean",
             The statistics upon which non-singleton i is thresholded.
             See scipy.cluster.hierarchy.fcluster for further details.
 
+         :param scoring_data: string or None
+            The form of the input data
```
glouppe: I would rephrase as "The type of input data to pass into the `scoring` function.`

beard/clustering/wrappers.py
```diff
@@ -79,17 +79,32 @@ def __init__(self, method="single", affinity="euclidean",
...
+            - "raw": Array of samples;
```
glouppe: I would say "original" array, to make things even clearer

beard/clustering/wrappers.py
```diff
@@ -121,25 +136,31 @@ def fit(self, X, y=None):
...
-        else:
+
+        elif self.scoring_data != "affinity":
             self.linkage_ = hac.linkage(X,
                                         method=self.method,
                                         metric=self.affinity)
+        else:
+            raise ValueError("Inputs combination error: Make sure to use\
+                a callable function for the affinity parameter. If you don't,\
+                you would need to set the scoring_data parameter to affinity.")
```
glouppe: I find the last `elif` statement difficult to understand. I would instead structure the code like this:

```
        Xr = X  # Xr as the original raw data
        Xa = None

        # Build linkage matrix
        if self.affinity == "precomputed":
            Xa = X  # Xa as the affinity matrix
            i, j = np.triu_indices(X.shape[0], k=1)
            X = X[i, j]
            self.linkage_ = hac.linkage(X, method=self.method)

        elif callable(self.affinity):
            X = self.affinity(X)
            Xa = X
            i, j = np.triu_indices(X.shape[0], k=1)
            X = X[i, j]
            self.linkage_ = hac.linkage(X, method=self.method)

        else:
            self.linkage_ = hac.linkage(X,
                                        method=self.method,
                                        metric=self.affinity)
```

and then below, when you call the scoring function, check if `scoring_data == "affinity" and Xa is None`, if which case, trigger the exception.

beard/clustering/wrappers.py
```diff
@@ -148,7 +169,7 @@ def fit(self, X, y=None):
         if y is not None:
             y_arr = np.array(y)
             all_y_neg = y_arr.sum() == len(y_arr) * -1
-            ground_truth = y is not None and not all_y_neg
+            ground_truth = not all_y_neg
```
glouppe: You can replace the whole thing with `ground_truth = np.any(np.array(y) != -1)`

beard/clustering/wrappers.py
```diff
@@ -148,7 +169,7 @@ def fit(self, X, y=None):
...
         else:
             ground_truth = False
```
glouppe: Or even better, the whole if-statement with `ground_truth = (y is not None) and np.any(np.array(y) != -1)

beard/clustering/tests/test_wrappers.py
glouppe: Can you please define the scoring functions directly within the test functions, as in the example above? It is easier to read and understand if everything is contained within the test

beard/clustering/tests/test_wrappers.py
```diff
@@ -27,91 +27,223 @@
...
     random_state = check_random_state(42)
 
     return make_blobs(centers=4, shuffle=False, random_state=random_state)
 
 
-def test_unsupervised(testData):
-    """Unsupervised clustering."""
-    X, y = testData
+@pytest.fixture
+def affinity_data():
```
glouppe: There is a lot of repeated code here. Cant we factor out the generation of data?

beard/clustering/wrappers.py
```diff
@@ -79,17 +79,32 @@ def __init__(self, method="single", affinity="euclidean",
             The statistics upon which non-singleton i is thresholded.
             See scipy.cluster.hierarchy.fcluster for further details.
 
+         :param scoring_data: string or None
+            The form of the input data
+            - "raw": Array of samples;
+            - "affinity": Affinity matrix.
```
glouppe: You should also specifiy that it can be `None`

beard/metrics/clustering.py
```diff
@@ -19,9 +20,28 @@
...
+    Please check sklearn.metrics.silhouette_score for the documentation
+    """
+    n_samples = X.shape[0]
+    n_clusters = len(np.unique(labels))
+    if 1 < n_clusters < n_samples:
+        return sklearn_silhouette_score(X, labels, metric)
+    else:
+        return -1
+
+
```
glouppe: Since you are adding the silhouette score in this PR, you could also add the others directly now I think.

beard/metrics/clustering.py
```diff
@@ -19,9 +20,28 @@
...
 
+def silhouette_score(X, labels, metric="precomputed"):
+    """Compute silhouette_score with an extension of number of clusters condition.
+
+    The condition is that Silhouette Coefficent is only defined if number of
+    clusters is 1 < n_clusters < n_samples. What the extension does here is
+    simply accepts n_clusters out of that boundary and returns the minimum
+    possible score for such cases.
+
+    Please check sklearn.metrics.silhouette_score for the documentation
```
glouppe: Please document the parameters

beard/clustering/wrappers.py
natsheh: It is mentioned 3 lines before :param scoring_data: string or None

glouppe: Yes, but it should that it that case, no data is passed

beard/metrics/clustering.py
natsheh: I would suggest keepingsilhouette only test case for this PR bug fix. The other would have a lot of code, for example Gap Statisctic PR that I am currently preparing. that would also need seperate test cases and pair reviewing. 
Another reason is that Gap Statistic, for example, doesn't provide a score value as in b3 and Silhouette, it searches for the best k number of clusters when it finds a statisticly big gap adding 1 more clusters. I think it will be too much for this bug fix PR, isn't it?

glouppe: yup ok fair enough

beard/metrics/clustering.py
natsheh: All done now I hope. Thanks @glouppe .

glouppe:
I fixed the Travis build. I will make some last cosmetic changes myself and then merge this afternoon.

natsheh:
Thanks! :)

glouppe:
Closed for #39

glouppe:
PR #39 is now merged. Thanks for the improvements!

PR #40, opened 2015-04-27
setup: include all packages

jalavik:
This PR fixes the installation of beard under `site-packages`.utils

Before
------

```shell
cd src/beard
pip install .
cd
ipython
```

```python
In [1]: from beard import utils
---------------------------------------------------------------------------
ImportError                               Traceback (most recent call last)
<ipython-input-1-7970c0420530> in <module>()
----> 1 from beard import utils

ImportError: cannot import name utils
```

After
-----

```shell
cd src/beard
pip install .
cd
ipython
```

```python
In [1]: from beard import utils

In [2]: 
```

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -7,4 +7,4 @@
 
 [run]
 source = beard
-omit = /*/tests/* /ext/*
+omit = /*/tests/*  */beard/ext/*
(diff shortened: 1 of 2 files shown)


glouppe:
Thanks for the fix! It has been merged as part of #41

PR #44, opened 2015-05-12
Adding supervised and unsupervised scoring function types

natsheh:


natsheh:
Currently, if we pass a supervised scoring function like b^3_F_score to the clusterer, the code will raise an error for the case of no ground truth (all y's are -1). This PR allows a proper usage of the clustering passing a supervised or unsupervised scoring function or both.  @MSusik @glouppe

beard/clustering/wrappers.py
```diff
@@ -36,8 +36,8 @@ class ScipyHierarchicalClustering(BaseEstimator, ClusterMixin):
 
     def __init__(self, method="single", affinity="euclidean",
                  threshold=None, n_clusters=None, criterion="distance",
-                 depth=2, R=None, monocrit=None, scoring=None,
-                 scoring_data=None):
+                 depth=2, R=None, monocrit=None, unsupervised_scoring=None,
+                 supervised_scoring=None, scoring_data=None):
```
MSusik: The only thing that puzzles me in this PR is the number of parameters passed to this constructor. A function that takes 11 arguments may be complicated to use.

natsheh: well, most of these parameters are inherited from the Scipy hierarchal clustering. The difference between the current status and this PR is one additional parameter that is necessary. This only 1 additional paramter will not only fix the current usage issue, but also will allow a hybrid usage of both supervised, when y is available and unsupervised when y is not available.

beard/clustering/wrappers.py
```diff
@@ -83,22 +83,27 @@ def __init__(self, method="single", affinity="euclidean",
...
-                    scoring(X, label_pred);
-                - Ground truth data and scoring_data == "affinity":
-                    scoring(Xa, label_true, label_pred);
-                - No ground truth data and scoring_data == "affinity":
-                    scoring(Xa, label_pred);
-                - Ground truth data and scoring_data is None:
-                    scoring(label_true, label_pred);
-                - No ground truth data and scoring_data is None:
-                    scoring(label_pred).
+            threshold. Ground truth must be True for this scoring function.
```
glouppe: " Ground truth must be True " does not actually mean anything. We should rather say: "If labels are provided in `y`, ..."

beard/clustering/wrappers.py
```diff
@@ -83,22 +83,27 @@ def __init__(self, method="single", affinity="euclidean",
...
+                - scoring_data == "raw":
+                    supervised_scoring(X_raw, label_true, label_pred);
+                - scoring_data == "affinity":
+                    supervised_scoring(X_affinity, label_true, label_pred);
+                - scoring_data is None:
+                    supervised_scoring(label_true, label_pred).
+
+        :param unsupervised_scoring: callable or None
+            The scoring function to maximize in order to estimate the best
+            threshold.  Ground truth must be False for this scoring function.
```
glouppe: Same here

beard/clustering/wrappers.py
```diff
@@ -163,11 +169,13 @@ def fit(self, X, y=None):
         # As default value we use the highest so we obtain only 1 cluster.
         best_threshold = self.linkage_[-1, 2]
         n_clusters = self.n_clusters
-        scoring = self.scoring
+        supervised_scoring = self.supervised_scoring
+        unsupervised_scoring = self.unsupervised_scoring
         threshold = self.threshold
         ground_truth = (y is not None) and np.any(np.array(y) != -1)
+        scoring = supervised_scoring is None and unsupervised_scoring is None
```
glouppe: Use the negative value. The name `scoring` suggests that it is true if there is a scoring function, while it is actually the opposite as you have stated it...

MSusik:
:+1:

glouppe:
Looks OK, merging

PR #46, opened 2015-05-21
general: no sudo with Travis

glouppe:
See http://blog.travis-ci.com/2015-03-31-docker-default-on-the-way/


glouppe:
This allows for shortened waiting times :)

PR #49, opened 2015-06-02
applications: initial release

MSusik:
This addresses #48. 

Signed-off-by: Mateusz Susik <mateusz.susik@cern.ch>

Issue #20, opened 2015-02-05
utils: better normalization of personal names


glouppe:
The current normalization of personal names introduced in #19 is working, but remains quite primitive and not very robust. 

A more elaborated schema might be needed. In any case, this should be checked on a larger set of personal name variants, so that most of them match upon normalization. 

(This is a difficult problem)

glouppe:
By looking at the results of author disambiguation, I realize that some obvious errors are due to LNFI blocking, which is currently based on a very simplistic normalization scheme for names. In particular, `normalize_name` dont remove name particles, even though we have code for that in `tokenize_name` (if  `drop_common_affixes=True`). 

@MSusik What would you think of adding a `drop_common_affixes=True` in `normalize_name`? 

Another systematic error I see is due to "ue" being a variant of "u" (e.g., Mueller, Muller) or "oe" being a variant of "o". I think adding a normalization step for this would improve our blocking without harming the clustering.

MSusik:
@glouppe 

> Another systematic error I see is due to "ue" being a variant of "u" (e.g., Mueller, Muller), or "oe" being a variant of "o", or "ae" being "a". I think adding a normalization step for this would improve our blocking without harming the clustering. 

It feels like implementing a metaphone algorithm from scratch. After we eliminate those three big cases, there will be still a lot of smaller ones to handle.

Still, I agree that removing common affixes will be an improvement.

glouppe:
Yes, true. 

+1 for adding `drop_common_affixes=True` only then.

kaplun:
Hi! I don't know enough to technical comment on this. Just wanted to remind the existence of a prior analysis that has been carried within INSPIRE related to name normalizations etc.

Might potentially save you many steps:
https://docs.google.com/document/d/1bd4y1WlgRDsEppBj4fejgT-FTMIHfQeHu0mrI-gOYt8/edit?usp=sharing

PR #52, opened 2015-06-11
applications: author disambiguation

glouppe:
Supersedes #49 with minor improvements.


glouppe:
Okay, I just checked that it still works on actual data. It does :) 

Further improvements regarding normalization (see #20) and blocking can come in subsequent PRs. 

Merging.

PR #58, opened 2015-06-12
utils: better name normalization (closes #20)

### LINKED ISSUE #20: utils: better normalization of personal names

MSusik:
TO DO:

- [x] score comparison ( before and after )

### DIFF
--- a/tests/utils/test_names.py
+++ b/tests/utils/test_names.py
@@ -33,7 +33,7 @@ def test_normalize_name():
     assert normalize_name("Doe, J.") == "doe j"
     assert normalize_name("Doe, J") == "doe j"
     assert normalize_name("Doe-Foe, Willem") == "doefoe willem"
-    assert normalize_name("Doe-Foe Willem") == "doe foe willem"
+    assert normalize_name("Doe-Foe Willem") == "willem doe foe"
     assert normalize_name("Dupont, René") == "dupont rene"
     assert normalize_name("Dupont., René") == "dupont rene"
     assert normalize_name("Dupont, Jean-René") == "dupont jean rene"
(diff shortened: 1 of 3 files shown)


tests/utils/test_names.py
```diff
@@ -43,6 +43,9 @@ def test_normalize_name():
     assert normalize_name("Dupont, J.-R.") == "dupont j r"
     assert normalize_name("Dupont") == "dupont"
     assert normalize_name("Dupont J.R.") == "dupont j r"
+    assert normalize_name("von und zu Hohenstein, F.") == "hohenstein f"
```
glouppe: Looks good overall, but could add a couple more tests for affixes? Also we should test when `drop_common_affixes=False`.

tests/utils/test_names.py
```diff
@@ -43,6 +43,20 @@ def test_normalize_name():
...
+    assert normalize_name("von und zu Hohenstein, F.",
+                          drop_common_affixes=False) == "vonundzuhohenstein f"
+    assert normalize_name("Jakub, Ibrahim ibn") == "jakub ibrahim ibn"
+    assert normalize_name("o'Neill, Jack") == "neill jack"
```
glouppe: Hmm, do you think it is in fact always a good idea to drop the affixes? E.g, it is very likely that we will have signatures with "Mac Donald" and "MacDonald". If we normalize like this, then these will end into two different blocks.

MSusik: That's why it is an argument to the function. Why I was implementing blocking algorithm I got higher recall when I dropped all the affixes. However, the strategy was different back then (double metaphone, moving signatures with multiple surnames to correct blocks, etc.) It might be worth to check the score after this change.

MSusik:
First result

528fd1ec491d55cfb9f0f63f4dfc83072fba4cfc

Number of blocks = 13114
True number of clusters 15575
Number of computed clusters 15755
B^3 F-score (overall) = 0.9814024143924055
B^3 F-score (train) = 0.9879790622231259
B^3 F-score (test) = 0.9810051562488993

Second result (with new name normalization) in progress.

natsheh:
:+1:

MSusik:
Results after cherry picking a38d632679c21524ec647d088f197adb4bc25e8b

I didn't resample the pairs.

Number of blocks = 13114
True number of clusters 15575
Number of computed clusters 15630
B^3 F-score (overall) = 0.9813912291609609
B^3 F-score (train) = 0.9878633340847559
B^3 F-score (test) = 0.981002451130001

MSusik:
After resampling
Number of blocks = 13114
True number of clusters 15575
Number of computed clusters 15649
B^3 F-score (overall) = 0.9815744332694658
B^3 F-score (train) = 0.9879289752654714
B^3 F-score (test) = 0.9811822815653505

So, it should be ready to merge!

natsheh:
:+1:

glouppe:
Old and new strategies give the exact same number of blocks, which is unexpected. Can you proofcheck the results?

MSusik:
OK, I rechecked it. Correct results:

('Number of blocks =', 13112)
('True number of clusters', 15575)
('Number of computed clusters', 16682)
('B^3 F-score (overall) =', 0.9832201238291286)

PR #61, opened 2015-06-17
examples: lighter coauthors extraction

MSusik:


MSusik:
Please note that it should merged no matter what is the difference in the scores, as with the current code the code crashes (segfault) on allocating a huge array in `SciPy`.

glouppe:
Merging, thanks

PR #64, opened 2015-06-24
matching: initial release

MSusik:
* Introduces a matching algorithm for the results of
  disambiguation. The results are matched to the disambiguation
  that was available previously, before the run of BEARD.

### DIFF
--- a/beard/matching/__init__.py
+++ b/beard/matching/__init__.py
@@ -1,0 +1,10 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of Beard.
+# Copyright (C) 2015 CERN.
+#
+# Beard is a free software; you can redistribute it and/or modify it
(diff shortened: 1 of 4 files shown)


glouppe:
For further reference (no need to take this into account in this PR), here is another approach I did using a linear programming formulation instead of the hungarian algorithm to solve the matching problem. The possibility to add random constraints yields more freedom for modeling the problem.
https://github.com/glouppe/notebooks/blob/master/One-side%20assignment%20problem%20with%20linear%20programming.ipynb

MSusik:
Your method probably should be added to the code as a replacement for munkres library

kaplun:
@glouppe what is the status of this branch?

glouppe:
Given the direction that @jacenkow has taken with invenio-beard, I think this branch will not be continued.

PR #65, opened 2015-07-09
clustering: new parallelization implementation

MSusik:
* Gets rid of Joblib

* Removes feature of passing negative value of n_jobs.

### DIFF
--- a/beard/clustering/blocking.py
+++ b/beard/clustering/blocking.py
@@ -16,10 +16,8 @@
 
 from __future__ import print_function
 
+import multiprocessing as mp
 import numpy as np
-
-from joblib import delayed
-from joblib import Parallel
 
 from sklearn.base import BaseEstimator
 from sklearn.base import clone
@@ -42,23 +40,49 @@ class _SingleClustering(BaseEstimator, ClusterMixin):
         return block_single(X)
 
 
-def _parallel_fit(fit_, partial_fit_, b, X, y, clusterer, verbose):
+def _parallel_fit(fit_, partial_fit_, estimator, verbose, data_queue,
+                  result_queue):
     """Run clusterer's fit function."""
-    if verbose > 1:
-        print("Clustering %d samples on block '%s'..." % (len(X), b))
+    # Status can be one of: 'middle', 'end'
+    # 'middle' means that there is a block to compute and the process should
+    # continue
+    # 'end' means that the process should finish as all the data was sent
+    # by the main process
+    status, block, existing_clusterer = data_queue.get()
 
-    if fit_ or not hasattr(clusterer, "partial_fit"):
-        try:
-            clusterer.fit(X, y=y)
-        except TypeError:
-            clusterer.fit(X)
-    elif partial_fit_:
-        try:
-            clusterer.partial_fit(X, y=y)
-        except TypeError:
-            clusterer.partial_fit(X)
+    if status == 'end':
+        # Special case where there are more processes than blocks
+        result_queue.put((None, None))
 
-    return (b, clusterer)
+    while status != 'end':
+
+        b, X, y = block
+
+        if len(X) == 1:
+            clusterer = _SingleClustering()
+        elif existing_clusterer and partial_fit_ and not fit_:
+            clusterer = existing_clusterer
+        else:
+            clusterer = clone(estimator)
+
+        if verbose > 1:
+            print("Clustering %d samples on block '%s'..." % (len(X), b))
+
+        if fit_ or not hasattr(clusterer, "partial_fit"):
+            try:
+                clusterer.fit(X, y=y)
+            except TypeError:
+                clusterer.fit(X)
+        elif partial_fit_:
+            try:
+                clusterer.partial_fit(X, y=y)
+            except TypeError:
+                clusterer.partial_fit(X)
+
+        result_queue.put((b, clusterer))
+        status, block, existing_clusterer = data_queue.get()
+
+    return
 
 
 class BlockClustering(BaseEstimator, ClusterMixin):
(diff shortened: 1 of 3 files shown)


beard/clustering/blocking.py
```diff
@@ -159,33 +183,48 @@ def _blocks(self, X, y, blocks):
...
-                   b, X_mask, y_mask, clusterer in self._blocks(X, y, blocks)))
-
-        for b, clusterer in results:
-            self.clusterers_[b] = clusterer
+        processes = []
+        # Here the blocks will be passed to subprocesses
+        data_queue = mp.Queue()
+        # Here the results will be passed back
+        result_queue = mp.Queue()
+        for x in range(0, self.n_jobs):
```
glouppe: `range(self.n_jobs)`

beard/clustering/blocking.py
```diff
@@ -159,33 +183,48 @@ def _blocks(self, X, y, blocks):
...
+            processes.append(mp.Process(target=_parallel_fit, args=(self.fit_,
+                             self.partial_fit_, self.base_estimator,
+                             self.verbose, data_queue, result_queue)))
+            processes[-1].start()
+
+        # First 16 blocks are sent into the queue without waiting for the
+        # results. This variable is a counter that takes care of this.
+        presend = 1
```
glouppe: I dont understand from the code how the first *16* are sent into the queue. Where does this number come from?

beard/clustering/blocking.py
```diff
@@ -159,33 +183,48 @@ def _blocks(self, X, y, blocks):
...
+            else:
+                presend += 1
+            if self.partial_fit_:
+                if block[0] in self.clusterers_:
+                    data_queue.put(('middle', block, self.clusterers_[b]))
+                    continue
+            data_queue.put(('middle', block, None))
+
+        # Get the last results and tell the subprocesses to finish
+        for x in range(0, self.n_jobs):
```
glouppe: `range(self.n_jobs)`

beard/clustering/blocking.py
```diff
@@ -159,33 +183,48 @@ def _blocks(self, X, y, blocks):
...
+                             self.partial_fit_, self.base_estimator,
+                             self.verbose, data_queue, result_queue)))
+            processes[-1].start()
+
+        # First 16 blocks are sent into the queue without waiting for the
+        # results. This variable is a counter that takes care of this.
+        presend = 1
+
+        for block in self._blocks(X, y, blocks):
+            if presend > self.n_jobs:
```
glouppe: From a rigorous loop invariant point of view, I would initialize `presend` to `0` and then check whether `presend >= self.n_jobs` instead. Or better, I would put this check after the `data_queue.put` call below (using a `>` then).

glouppe:
Besides my nitpicks, this looks good to me. Thanks! Two questions though:

- Does this solve the blocking issue we had?
- Have you checked this works fine both for Python 2.7 and 3.4?

MSusik:
> Does this solve the blocking issue we had?

Yes, just a one small improvement is needed, as this implementation can hang.

> Have you checked this works fine both for Python 2.7 and 3.4?

Yes.

glouppe:
Great then, +1 for merge once my comments are fixed.

glouppe:
CC: @ogrisel Just to let you know we have add hanging issues with joblib -- processes are all stalling, CPU usage goes down to 0, and then nothing more happens. Do you know where this could be coming from? These are very difficult to reproduce and seem to appear at random... Directly using `multiprocessing` solves our immediate problem, but it would be nice if joblib could be used again.

MSusik:
Note that the approach is completely different as joblib spawns a process for every block of data, while I keep a pool of processes that run all the time.

ogrisel:
I want to make it possible to reuse joblib pools too across several consecutive calls to `Parallel.__call__`. However this is probably not the cause of the hanging.

Which implementation of BLAS do you use when you observe the hanging? anaconda's MKL? OSX's builtin accelerate?

ogrisel:
> Note that the approach is completely different as joblib spawns a process for every block of data, while I keep a pool of processes that run all the time.

Note: joblib spawns a pool with a fixed number of worker process per call to Parallel (using a multiprocessing `Pool` instance using the the `apply_async` under the hood). But you can pass many blocks of data and the number of workers stay constants.

ogrisel:
BTW, could you also tell be if you observe the hanging behavior when enabling the `forkserver` start method under Python 3.4+? To try that you need to modify the main block of the main script that starts your Python program with:

```
import multiprocessing as mp
# import your modules here

if __name__ == '__main__':
    mp.set_start_method('forkserver')
    # call your code here
```

MSusik:
> Which implementation of BLAS do you use when you observe the hanging? anaconda's MKL?

We observed the error when using anaconda's MKL and when working outside of anaconda. For example here is a numpy's config for the run without anaconda:

```
blas_info:
    libraries = ['blas']
    library_dirs = ['/usr/lib']
    language = f77
lapack_info:
    libraries = ['lapack']
    library_dirs = ['/usr/lib']
    language = f77
atlas_threads_info:
  NOT AVAILABLE
blas_opt_info:
    libraries = ['blas']
    library_dirs = ['/usr/lib']
    language = f77
    define_macros = [('NO_ATLAS_INFO', 1)]
atlas_blas_threads_info:
  NOT AVAILABLE
openblas_info:
  NOT AVAILABLE
lapack_opt_info:
    libraries = ['lapack', 'blas']
    library_dirs = ['/usr/lib']
    language = f77
    define_macros = [('NO_ATLAS_INFO', 1)]
atlas_info:
  NOT AVAILABLE
lapack_mkl_info:
  NOT AVAILABLE
blas_mkl_info:
  NOT AVAILABLE
atlas_blas_info:
  NOT AVAILABLE
mkl_info:
  NOT AVAILABLE
```

ogrisel:
That's weird. Do you use OpenMP-based libraries or compiled extensions (e.g. Cython `prange` constructs)?

Would be great to provide a standalone joblib snippet that reproduce the freeze so that I can try to debug.

PR #67, opened 2015-07-16
clustering: better way to finish processes.

MSusik:


MSusik:
In the previous PR I assumed that multiprocessing queues are FIFO. Unfortunately it's not true, and in the previous implementation some subprocesses  might not finish.

PR #70, opened 2015-07-17
Adding the ethnicity estimator to the features

natsheh:


examples/applications/author-disambiguation/distance.py
```diff
@@ -13,13 +13,16 @@
 
 .. codeauthor:: Gilles Louppe <g.louppe@cern.ch>
 .. codeauthor:: Mateusz Susik <mateusz.susik@cern.ch>
+.. codeauthor:: Hussein AL-NATSHEH <hussein.al.natsheh@cern.ch>
 
 """
 
 import argparse
 import pickle
+import cPickle
```
MSusik: not available in python3.

Anyway, if we want to use this race estimator in Python3 as well, we will have to recreate the file (behaviour of pickle depends on which version of Python do you use).

glouppe:
Closing for #75

PR #73, opened 2015-07-29
script for building a balanced pair dataset

MSusik:


MSusik:
@glouppe Ready to review.

examples/applications/author-disambiguation/sampling.py
```diff
@@ -0,0 +1,255 @@
...
+# This file is part of Beard.
+# Copyright (C) 2015 CERN.
+#
+# Beard is a free software; you can redistribute it and/or modify it
+# under the terms of the Revised BSD License; see LICENSE file for
+# more details.
+
+"""Script for generating the training set.
+
+It samples pairs of signatures labled with 1 if they are of different authors
```
glouppe: labeled

examples/applications/author-disambiguation/sampling.py
```diff
@@ -0,0 +1,255 @@
...
+or 0 if they are of the same author.
+
```
glouppe: It would help also to add an example of line command usage.

examples/applications/author-disambiguation/sampling.py
```diff
@@ -0,0 +1,255 @@
...
+        It can only be:
+        -   None; if the blocking function is block_last_name_first_initial
+        -   int; if the blocking function is block_double_metaphone
+            please check the documentation of double_metaphone blocking in
+            beard.clustering.blocking_funcs.py
+
+    :param clusters_filename: string
+        Path to the input clusters (ground-truth) file
+
+    :param test_size: float
```
glouppe: remove this?

examples/applications/author-disambiguation/sampling.py
```diff
@@ -0,0 +1,255 @@
...
+
+from __future__ import print_function
+
+import argparse
+import json
+import math
+import numpy as np
+import random
+
+from sklearn.cross_validation import train_test_split
```
glouppe: this is no longer used

examples/applications/author-disambiguation/sampling.py
```diff
@@ -0,0 +1,255 @@
...
+    This function builds a pair dataset from claimed signatures.
+    It gives the ability to specify the
+    blocking function and whether the sampling would be balanced or not.
+
+    Parameters
+    ----------
+    :param blocking_function: string
+        must be a defined blocking function. Defined functions are:
+        - block_last_name_first_initial
+        - block_double_metaphone
```
glouppe: Use quotes around the possible values?

examples/applications/author-disambiguation/sampling.py
```diff
@@ -0,0 +1,255 @@
...
+
+    :param clusters_filename: string
+        Path to the input clusters (ground-truth) file
+
+    :param test_size: float
+        The size of the test set: 0.0 < size < 1.0.
+        It is used for train_test_split. Should be the same as in clustering
+
+    :param balanced: boolean
+        determines if the sampling would be balanced or uniformly random
```
glouppe: we should be more explicit by what is meant by "balanced"

examples/applications/author-disambiguation/sampling.py
```diff
@@ -0,0 +1,255 @@
...
+from beard.clustering import block_last_name_first_initial
+
+
+def pair_sampling(blocking_function,
+                  blocking_threshold,
+                  clusters_filename,
+                  balanced, verbose,
+                  train_filename,
+                  sample_size,
+                  use_blocking):
```
glouppe: Can we set good default values to balanced, verbose, sample_size and use_blocking?

examples/applications/author-disambiguation/sampling.py
```diff
@@ -0,0 +1,255 @@
...
+        list of signature pairs
+    """
+    # Load ground-truth
+    true_clusters = json.load(open(clusters_filename, "r"))
+    clusters_reversed = {v: k for k, va in true_clusters.iteritems()
+                         for v in va}
+
+    train_signatures = json.load(open(train_filename, "r"))
+
+    if not use_blocking:
```
glouppe: This code is every monolithic and only specifc to `use_blocking=False`. (and the other part specific to the case when it is True). Dont you think we should split this function into two sub private functions?

examples/applications/author-disambiguation/sampling.py
glouppe: also, we should indicate that this only works if `use_blocking=True`

glouppe:
Except my comment to split the code, +1 for merge 

(even if this is not the most polished code ever, let us not waste too long on this)

MSusik:
Samping without blocking part extracted to another function. Can be merged.

PR #84, opened 2016-01-21
blocking: multiprocessing

jacenkow:
* Removes multiprocessing for n_jobs = 1.

Signed-off-by: Grzegorz Jacenków <grzegorz.jacenkow@cern.ch>

### DIFF
--- a/beard/clustering/blocking.py
+++ b/beard/clustering/blocking.py
@@ -16,7 +16,6 @@
 
 from __future__ import print_function
 
-import multiprocessing as mp
 import numpy as np
 import time
 
@@ -26,11 +25,6 @@ from sklearn.base import ClusterMixin
 from sklearn.utils import column_or_1d
 
 from .blocking_funcs import block_single
-
-try:
-    from multiprocessing import SimpleQueue
-except ImportError:
-    from multiprocessing.queues import SimpleQueue
 
 
 class _SingleClustering(BaseEstimator, ClusterMixin):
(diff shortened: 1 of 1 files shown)


jacenkow:
@glouppe

beard/clustering/blocking.py
```diff
@@ -191,51 +214,82 @@ def _fit(self, X, y, blocks):
...
-                blocks_computed += 1
+        if self.n_jobs == 1:
+            blocks_computed = 0
+            blocks_all = len(np.unique(blocks))
+
+            for block in self._blocks(X, y, blocks):
+                if self.partial_fit_ and block[0] in self.clusterers_:
+                    data = (block, self.clusterers_[block[0]])
+
+                data = (block, None)
```
glouppe: this should in an `else` clause of the previous if-statement

beard/clustering/blocking.py
```diff
@@ -88,6 +82,35 @@ def _parallel_fit(fit_, partial_fit_, estimator, verbose, data_queue,
...
+            clusterer.fit(X, y=y)
+        except TypeError:
+            clusterer.fit(X)
+    elif partial_fit_:
+        try:
+            clusterer.partial_fit(X, y=y)
+        except TypeError:
+            clusterer.partial_fit(X)
+
+    return (b, clusterer)
```
glouppe: Obviously, some code could be factored out of `_single_fit` and `_parallel_fit`, but that is a minor issue.

MSusik:
>Obviously, some code could be factored out of _single_fit and _parallel_fit, but that is a minor issue

So, what about the approach that Gilles suggested earlier?

```python
if self.n_jobs == 1:
    from Queue import Queue
    data_queue = Queue()
    result_queue = Queue()
    _parallel_fit(...)
else:
    try:
        from multiprocessing import SimpleQueue
    except ImportError:
        from multiprocessing.queues import SimpleQueue
    data_queue = SimpleQueue()
    result_queue = SimpleQueue()
    # The rest of the code
```

I think `Queue` should not copy the data, only the reference, so memory consumption shouldn't be a problem.

jacenkow:
@MSusik ```_parallel_fit``` gets an empty queue (with no data), thus goes into an infinite loop.

MSusik:
@jacenkow You're right. It obviously runs in the same thread :stuck_out_tongue_closed_eyes:

jacenkow:
@MSusik haha I have done exactly in the same way as you did, and to be honest all the credits for spotting the problem goes to Gilles :dancer:

MSusik:
Sidenote: You still should be able to implement this solution by using Python threading which doesn't spawn any processes, though.

Issue #87, opened 2016-03-22
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 9: ordinal not in range(128)


lfoppiano:
Dear all, 
   after reading the paper and find it very interesting I wanted to try the application, so I cloned the repository. I've run the setup (using virtualenv) and I launched 

`python author_disambiguation.py`

But then I get the following error: 
```
Traceback (most recent call last):
  File "/Users/lfoppiano/development/inria/inria-virtualenv/lib/python3.5/site-packages/numpy/lib/format.py", line 638, in read_array
    array = pickle.load(fp, **pickle_kwargs)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 9: ordinal not in range(128)

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/lfoppiano/development/inria/disambiguation/paper-author-disambiguation/beard/examples/author_disambiguation.py", line 74, in <module>
    X = data["X"]
  File "/Users/lfoppiano/development/inria/inria-virtualenv/lib/python3.5/site-packages/numpy/lib/npyio.py", line 224, in __getitem__
    pickle_kwargs=self.pickle_kwargs)
  File "/Users/lfoppiano/development/inria/inria-virtualenv/lib/python3.5/site-packages/numpy/lib/format.py", line 644, in read_array
    "to numpy.load" % (err,))
UnicodeError: Unpickling a python object failed: UnicodeDecodeError('ascii', b'Salt, Jos\xc3\xa9', 9, 10, 'ordinal not in range(128)')
You may need to pass the encoding= option to numpy.load
```
I've tried to use different encodings when loading the file 
`data = np.load("data/author-disambiguation.npz")`

but with no success. 

What am I missing?

glouppe:
Hi! Can you try the following:

```data = np.load("data/author-disambiguation.npz", encoding="latin1")```

natsheh:
@lfoppiano I recommend trying out the example here:
https://github.com/inspirehep/beard/tree/master/examples/applications/author-disambiguation

lfoppiano:
@glouppe I've tried also that and I've got another error: 
`Traceback (most recent call last):
  File "/Users/lfoppiano/development/inria/disambiguation/paper-author-disambiguation/beard/examples/author_disambiguation.py", line 86, in <module>
    block_clusterer.fit(X)
[...]

Issue #80, opened 2015-11-09
Annoying install issue between scipy and scikit-learn


jacquerie:
When running
```shell
(beard) $  pip install --no-cache-dir git+https://github.com/inveniosoftware/beard@master#egg=beard
```
in a fresh virtualenv we get:
```shell
Collecting beard from git+https://github.com/inveniosoftware/beard@master#egg=beard
  Cloning https://github.com/inveniosoftware/beard (to master) to /tmp/pip-build-FTsam1/beard
Collecting jellyfish (from beard)
  Downloading jellyfish-0.5.1.tar.gz
Collecting numpy>=1.9 (from beard)
  Downloading numpy-1.10.1.tar.gz (4.0MB)
    100% |████████████████████████████████| 4.1MB 55.5MB/s 
Collecting scikit-learn>=0.15.2 (from beard)
  Downloading scikit-learn-0.17.tar.gz (7.8MB)
    100% |████████████████████████████████| 7.8MB 29.8MB/s 
Collecting scipy>=0.14 (from beard)
  Downloading scipy-0.16.1.tar.gz (12.2MB)
    100% |████████████████████████████████| 12.2MB 57.4MB/s 
Collecting six (from beard)
  Downloading six-1.10.0-py2.py3-none-any.whl
Collecting unidecode (from beard)
  Downloading Unidecode-0.04.18.tar.gz (206kB)
    100% |████████████████████████████████| 208kB 55.8MB/s 
Collecting fuzzy (from beard)
  Downloading Fuzzy-1.1.tar.gz
Installing collected packages: jellyfish, numpy, scikit-learn, scipy, six, unidecode, fuzzy, beard
  Running setup.py install for jellyfish
  Running setup.py install for numpy
  Running setup.py install for scikit-learn
    Complete output from command /home/jnotarst/.virtualenvs/beard/bin/python -c "import setuptools, tokenize;__file__='/tmp/pip-build-FTsam1/scikit-learn/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-4HmVs6-record/install-record.txt --single-version-externally-managed --compile --install-headers /home/jnotarst/.virtualenvs/beard/include/site/python2.7/scikit-learn:
    Partial import of sklearn during the build process.
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
[...]

PR #89, opened 2016-03-24
Example documentation update

lfoppiano:
Hi, 
 I have added some information in the example's directories on how to build it. 

Cheers
Luca


MSusik:
Thank you for your PR.

What's the reason for leaving the `.rst` format in favour of `.md`?

lfoppiano:
I've could not use the github markup, with links  and so on, so I've opted to rename them .md.

examples/README.md
```diff
@@ -0,0 +1,31 @@
+# Examples
+
+General purpose and introductory examples of Beard.
+
+## Getting started
+
+This application works only on Python 2.x. On Linux should be already installed.
```
MSusik: That's simply not true.

Please see the build: https://travis-ci.org/inspirehep/beard/jobs/118296657
Python 3.x works as well.

MSusik:
> I've could not use the github markup, with links and so on, so I've opted to rename them .md. 

We prefer `.rst` as it is richer.

lfoppiano:
Indeed it builds, but I could not manage to run it (see #87 ).

MSusik:
OK, now I understand.

Basically, the library works for `Python3`, but the examples are not Python3 compatible. I will fix it soon as this might cause the problems that you already had.

examples/README.md
```diff
@@ -0,0 +1,31 @@
...
+cd virtualenv
+source bin/activate
+```
+
+...and install all the required libraries: 
+
+```
+pip install numpy 
+pip install scipy
+```
```
MSusik: Actually `pytho2n setup.py install` should take care of this.

See https://github.com/inspirehep/beard/blob/master/setup.py#L87

lfoppiano:
Indeed. :+1:

MSusik:
https://github.com/inspirehep/beard/pull/90 should fix Python3 compability

> Regarding the libraries, yes, but I had to install them manually, because the setup was requesting them. 

Can you tell me more in detail what happened? `python setup.py install` should install dependencies automatically for you.

MSusik:
Inactive

lfoppiano:
Sorry, I missed your question. The setup was stopping saying that there was no numpy library.

jacquerie:
Yeah, @kaplun, this is the problem I was talking about. Since `beard` depends on `scikit-learn` we're going to run in variations of #80 if you first don't install `numpy` and `scipy`. I think we should point to http://scikit-learn.org/stable/install.html in the installation instructions.

MSusik:
@lfoppiano I managed to reproduce your problem:

```
ImportError: Numerical Python (NumPy) is not installed.
scikit-learn requires NumPy >= 1.6.1.
Installation instructions are available on the scikit-learn website: http://scikit-learn.org/stable/install.html
```
It seems that scikit-learn doesn't install its numpy requirement. We'll see how we can solve this problem.

kaplun:
IIRC `scikit-learn` doesn't install any `numpy` in order not to force installation of an under-performant flavor (since the preferred way is through Anaconda).

So probably there is no automatic solution to our problem?

glouppe:
Since both numpy and scipy are common requirements of scikit-learn and beard, how could this be an issue? @MSusik Do you understand why `setup.py` does not automatically take of this?

MSusik:
Sorry for the delay

@glouppe Apparently there is no way of ensuring that the dependencies in `install_requires` are installed
in a desired order. Somehow the dependency resolving algorithm in pip seems to even prioritize installing
`scikit-learn` in front of `numpy`.

One might think that it should be possible to fix this by adding the more important dependencies to `setup_requires` - a field that lists all the dependencies that should be installed before anything else
happens in `setup.py`. Unfortunately packages installed like this are not available for the packages
listed in `install_requires` :angry: 
https://pythonhosted.org/setuptools/setuptools.html#new-and-changed-setup-keywords

Maybe we should approach the problem from the other side. We all know that installing beard with
its all dependencies with pip is not the desired solution. Perhaps we should throw a warning
stating that firstly the user should install conda. Or throw a warning that a good numpy installation
is required before continuing with `setup.py`.

lfoppiano:
Guys, from my point o view (as a user) you could just add a note in the documentation ;-) and that plus the message I would eventually get, will do. 

My 2 cents. 
Luca

PR #92, opened 2016-04-21
similarity: threshold transformer

MSusik:
Signed-off-by: Mateusz Susik <mateuszsusik@gmail.com>

### DIFF
--- a/beard/similarity/pairs.py
+++ b/beard/similarity/pairs.py
@@ -330,6 +330,51 @@ class ElementMultiplication(TransformerMixin):
         return Xt
 
 
+class Thresholder(TransformerMixin):
+    """Element-wise multiplication on paired data."""
+
+    def __init__(self, threshold):
+        """Initialize.
+
+        Parameters
+        ----------
+        :param threshold: The threshold over which the transformer will output
+            1.
+        """
+        self.threshold = threshold
+
+    def fit(self, X, y=None):
+        """(Do nothing).
+
+        Parameters
+        ----------
+        :param X: array-like, shape (n_samples, n_features)
+            Input data.
+
+        Returns
+        -------
+        :returns: self
+        """
+        return self
+
+    def transform(self, X):
+        """Return 1 if value exceeds the threshold.
+
+        0 otherwise
+
+        Parameters
+        ----------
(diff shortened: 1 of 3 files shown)


MSusik:
Review from email by @natsheh 

> After a quick check, all seems to be fine.

Merging.

beard/similarity/pairs.py
```diff
@@ -330,6 +330,51 @@ def transform(self, X):
         return Xt
 
 
+class Thresholder(TransformerMixin):
+    """Element-wise multiplication on paired data."""
```
natsheh: Need to change this doc

examples/applications/author-disambiguation/distance.py
```diff
@@ -263,20 +264,21 @@ def _build_distance_estimator(X, y, verbose=0, ethnicity_estimator=None,
                 ("classifier", EstimatorTransformer(ethnicity_estimator)),
             ]), groupby=group_by_signature)),
             ("sigmoid", FuncTransformer(func=expit)),
-            ("combiner", ElementMultiplication())
+            ("combiner", ElementMultiplication()),
+            ("thresholder", Thresholder(0.25))
```
natsheh: better to pass the threshold value as an argument

MSusik: @natsheh Do you mean that the threshold should be passed from the command line?

kaplun:
Hi! What are you working on? What impact is it going to have on our usage at INSPIRE?

examples/applications/author-disambiguation/distance.py
natsheh: yes, and the 0.25 could be the default

MSusik:
@kaplun We're adjusting one thing for our publication. Specifically, we're changing the way how the ethnicity features are used. These changes will be probably reverted, and if not, they won't have much impact at INSPIRE (we will leave them if they improve the accuracy). Anyway, I believe you won't be using ethnicity features for the disambiguation at INSPIRE, so you won't notice any difference.

Sorry for some glitches, like no commit log, but we're a but in a hurry (you know, deadlines :smile:)

PR #93, opened 2016-04-23
examples: thresholder removal

MSusik:
As the Thresholder is not longer needed, it is removed from the exemplary disambiguation.

Signed-off-by: Mateusz Susik <mateuszsusik@gmail.com>


MSusik:
Reverting part of the changes done in #92

PR #98, opened 2018-03-13
sampling is fixed to run on python3 and a bug in category_size variable is fixed

hajianOne:


MSusik:
@hajianOne 
I see some troubles with Travis. Will merge it once I fix them.

Thanks for the PR!

PR #104, opened 2020-07-30
clustering: add logging

MJedr:


pazembrz:
Don't you prefer to use f-strings? By default we are using them in all of our python3 projects.

MJedr:
I prefer, but python 2 doesn't :P

